PR #77, opened 2019-01-11
Fix getting started example (AttributeError: 'ItemCountMonitor' object has no attribute 'stats')

cuducos:
When running the first monitor example from the Getting Started chapter, users would eventually get an error such as:

```
AttributeError: 'ItemCountMonitor' object has no attribute 'stats'
```

Arguably this could be fixed by replacing `self.stats` by `self.data.stats` or by using the `StatsMonitorMixin`. As this is a getting started section, this commit opted for the simpler solution (less objects to import, more explicit API).

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -94,8 +94,7 @@ your monitors.
 
         @monitors.name('Minimum number of items')
         def test_minimum_number_of_items(self):
-            item_extracted = getattr(
-                self.stats, 'item_scraped_count', 0)
+            item_extracted = self.data.stats.get('item_scraped_count', 0)
             minimum_threshold = 10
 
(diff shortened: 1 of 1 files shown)


rennerocha:
The sample code in `examples/tutorial` relies on `StatsMonitorMixin`. Would you mind to update it too, so the docs and the example code are in sync?

cuducos:
> Would you mind to update it too, so the docs and the example code are in sync?

Actually I do hahaha… as I argued:

> As this is a getting started section, this commit opted for the simpler solution (less objects to import, more explicit API).

Thus I just would like to double-check: the maintainability issue (making sure _the docs and the example code are in sync_) is more important at this point than being beginner friendly? If so, I update ; )

rennerocha:
It should be beginner friendly AND in sync :-)

Issue #73, opened 2019-01-10
Update the "Readme.md" to reflect new links and installation instructions


raphapassini:
We should have a link to the complete documentation that we already have and it's publish in readthedocs
and also put some installation notes in the readme.

PR #81, opened 2019-01-15
Suggest PyPI as recommended installation source

jesuslosada:
Suggest a stable PyPI release as default installation source instead of the git repo.


rennerocha:
Related to #73 .

PR #83, opened 2019-01-22
Sentry integration

muzaffaryousaf:
Add sentry action to spidermon so that the notifications could be sent to sentry.


spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import absolute_import
+from __future__ import print_function
+
+from spidermon.contrib.actions.templates import ActionWithTemplates
+from spidermon.exceptions import NotConfigured
+
+
+class SendMessage(ActionWithTemplates):
```
rennerocha: This action doesn't render any template, so it doesn't need to inherit from it. Is there any other reason to use this inheritance?

spidermon/contrib/actions/sentry/sentry.py
```diff
@@ -0,0 +1,46 @@
...
+            scope.set_extra("spider_name", message.get('spider_name', ''))
+            scope.set_extra("spider_stats", message.get('spider_stats', {}))
+            scope.set_extra("items_count", message.get('items_count', 0))
+
+            scope.set_extra("passed_monitors_count", message.get('passed_monitors_count', 0))
+            scope.set_extra("failed_monitors_count", message.get('failed_monitors_count', 0))
+
+            scope.set_extra("failed_monitors", message.get('failed_monitors', []))
+
+            sentry_client.capture_event({"message": message.get('title'), 'level': 'error'}, scope=scope)
```
rennerocha: Why did you fixed the level of the message as an error? What if I want to register the result of my spider even if there is no errors?

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+        self.title = title or self.title
+        self.fake = fake or self.fake
+
+        if not self.title:
+            raise NotConfigured("You must provide a title for the message.")
+
+    @classmethod
+    def from_crawler_kwargs(cls, crawler):
+        return {
+            'title': crawler.settings.get('SPIDERMON_MESSAGE_TITLE'),
```
rennerocha: What do you think about `SPIDERMON_SENTRY_MSG_TITLE` to be more descriptive?

spidermon/contrib/actions/sentry/sentry.py
```diff
@@ -0,0 +1,46 @@
...
+
+from sentry_sdk.client import Client
+from sentry_sdk import configure_scope
+
+
+class SendSentryMessage(SendMessage):
+    def __init__(self, sentry_log_level=None, sentry_dsn=None, sentry_enabled=None, *args, **kwargs):
+        super(SendSentryMessage, self).__init__(*args, **kwargs)
+
+        self.sentry_log_level = sentry_log_level or self.sentry_log_level
```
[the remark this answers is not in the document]
muzaffaryousaf: This was supposed to use in send message.  I forgot to give it final look before asking for review.

spidermon/contrib/actions/sentry/__init__.py
muzaffaryousaf: The actual reason is to get context data which is being sent to these templates.

spidermon/contrib/actions/sentry/sentry.py
muzaffaryousaf: Going to pass the value of SPIDERMON_SENTRY_LOG_LEVEL

spidermon/contrib/actions/sentry/sentry.py
```diff
@@ -0,0 +1,46 @@
...
+        kwargs = super(SendSentryMessage, cls).from_crawler_kwargs(crawler)
+        kwargs.update({
+            'sentry_dsn': crawler.settings.get('SENTRY_DSN'),
+            'sentry_enabled': crawler.settings.get('SENTRY_ENABLED'),
+            'sentry_log_level': crawler.settings.get('SENTRY_LOG_LEVEL'),
+        })
+        return kwargs
+
+    def send_message(self, message):
+        if not self.sentry_enabled:
```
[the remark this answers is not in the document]
muzaffaryousaf: Sentry can emit event automatically for any error via logs. I think since we are not using that feature i'll remove it from here.

muzaffaryousaf:
@rennerocha It's ready for 2nd iteration.

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_message(self):
+        """
+        Returns message dictionary
+        """
+        message = dict()
+
+        message['title'] = self.get_title()
+
+        if self.data.job:
+            message['job_link'] = "https://app.scrapinghub.com /p/{job_id}".format(job_id=self.data.job.key)
```
rennerocha: - This line is too long (109 character).
- It also has an extra space after `.com`
- We have mixed single-quotes and double-quotes.

spidermon/contrib/actions/sentry/sentry.py
```diff
@@ -0,0 +1,42 @@
...
+            scope.set_extra("spider_name", message.get('spider_name', ''))
+            scope.set_extra("spider_stats", message.get('spider_stats', {}))
+            scope.set_extra("items_count", message.get('items_count', 0))
+
+            scope.set_extra("passed_monitors_count", message.get('passed_monitors_count', 0))
+            scope.set_extra("failed_monitors_count", message.get('failed_monitors_count', 0))
+
+            scope.set_extra("failed_monitors", message.get('failed_monitors', []))
+
+            sentry_client.capture_event({"message": message.get('title'), 'level': self.sentry_log_level}, scope=scope)
```
rennerocha: This line is too long (120 chars). I believe it would be good to keep the line size close to PEP8 suggestion of 80 chars.

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+            'title': crawler.settings.get('SPIDERMON_SENTRY_MESSAGE_TITLE'),
+            'fake': crawler.settings.getbool('SPIDERMON_SENTRY_FAKE'),
+        }
+
+    def run_action(self):
+        message = self.get_message()
+        if not self.fake:
+            self.send_message(message)
+        else:
+            print('-'*40)
```
rennerocha: Logging with `print` doesn't look good. What do you think include a real logger in this action?

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+            print(message)
+            print('-'*40)
+
+    def get_title(self):
+        if self.title:
+            return self.render_text_template(self.title)
```
rennerocha: This value seems to be always a simple string. Is it really necessary to render it from a template?

Maybe it would be even more clear to remove this method and change L51 to:
`message['title'] = self.title or 'Report Title`

spidermon/contrib/actions/sentry/__init__.py
rennerocha: The context used in `get_message` is provided by any `Action`, not only `ActionWithTemplates` (`self.result` and `self.data` contains everything needed). For me, the title rendering method is unnecessary, as it is a simple string. IMHO, inheriting only from `Action` will make the action simpler, as other user will not struggle trying to find ways to use templates.

spidermon/contrib/actions/sentry/sentry.py
```diff
@@ -0,0 +1,42 @@
+from spidermon.contrib.actions.sentry import SendMessage
+from spidermon.exceptions import NotConfigured
+
+from sentry_sdk.client import Client
+from sentry_sdk import configure_scope
+
+
+class SendSentryMessage(SendMessage):
```
rennerocha: Is it really necessary to have two classes to implement this action (`SendSentryMessage` and `SendMessage`)? All methods in `SendMessage` looks much coupled with `SendSentryMessage`, so maybe a single class would be enough.

spidermon/contrib/actions/sentry/__init__.py
muzaffaryousaf: I'm considering line length to 120 characters to keep it more readable. Also, i see the same length in this directory.

spidermon/contrib/actions/sentry/__init__.py
muzaffaryousaf: Just trying to be consistent what we have for exisitng classes.

spidermon/contrib/actions/sentry/__init__.py
muzaffaryousaf: We are meant to replace few variables in this string so i would say yes let's do it via the mechanism that we have already.

spidermon/contrib/actions/sentry/sentry.py
muzaffaryousaf: Just trying to be consistent here as well. We are not sure if somebody wants to extend the behaviour in future that's why i prefered to keep it similar to existing actions.

muzaffaryousaf:
> Included more comments. Apparently the interface for this action will not change, then you could start including the docs of this feature in the PR, so we merge everything together.

@rennerocha We want this to get merge ASAP because we are blocked on this, adding documentation will keep it open for few more days which we don't want. I've already added a issue for this and will create separate PR for that.

muzaffaryousaf:
@rennerocha This is ready for final review, please let me know when you are done with this. Thanks.

manycoding:
> > Included more comments. Apparently the interface for this action will not change, then you could start including the docs of this feature in the PR, so we merge everything together.
> 
> @rennerocha We want this to get merge ASAP because we are blocked on this, adding documentation will keep it open for few more days which we don't want. I've already added a [issue](https://github.com/scrapinghub/spidermon/issues/86) for this and will create separate PR for that.

Documentation in this case is essential since you added a new feature. Until this pr is not merged you can make a fork.

spidermon/contrib/actions/sentry/__init__.py
rennerocha: 120 character is fine. But there is still an extra space after `.com`

spidermon/contrib/actions/sentry/__init__.py
rennerocha: This is not good in the other classes too, so I believe we could start fixing it in this class now, and then use it as a pattern to fix other classes.

rennerocha:
> > Included more comments. Apparently the interface for this action will not change, then you could start including the docs of this feature in the PR, so we merge everything together.
> 
> @rennerocha We want this to get merge ASAP because we are blocked on this, adding documentation will keep it open for few more days which we don't want. I've already added a [issue](https://github.com/scrapinghub/spidermon/issues/86) for this and will create separate PR for that.

This can't be merged without documentation. We are including a new feature and documentation should be treated as part of the development process. If you need this in your project right now, you can use your branch.

muzaffaryousaf:
> > @rennerocha We want this to get merge ASAP because we are blocked on this, adding documentation will keep it open for few more days which we don't want. I've already added a [issue](https://github.com/scrapinghub/spidermon/issues/86) for this and will create separate PR for that.

@manycoding Thanks for your input. yeah, there is no doubt in having a documentation for any feature but only in different pull requests. Obviously, we are not going to release it without documentation.

spidermon/contrib/actions/sentry/__init__.py
raphapassini: +1 for having a logging. Print statements as a log is not a good idea. Let's starting fixing it in everything new we add, latter we can refactory the old code.

spidermon/contrib/actions/sentry/__init__.py
raphapassini: I agree with @muzaffaryousaf on this one, make it extensible it's better. Maybe you want to put the spidername or customer name in the title in a near future.

raphapassini:
Documentation needs to be part of this PR, otherwise we are adding code without any documentation, what about if the documentation PR can be done only after 3 months it get's merged?

spidermon/contrib/actions/sentry/__init__.py
rennerocha: I am still not convinced that the title will be rendered by a template (and inherited from `ActionWithTemplates`), but if we want to follow the pattern available in other similar actions, it would be interesting to include a new setting to allow the user easily provide custom context in the action as:

- https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/email/__init__.py#L69
- https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/reports/__init__.py#L21

spidermon/contrib/actions/sentry/__init__.py
muzaffaryousaf: i've already removed the extra space, please check in the diff, this view is outdated.

spidermon/contrib/actions/sentry/sentry.py
rennerocha: If they want to extend the functionality, they could just extend a unique `SendSentryMessage`. IMHO to have a separate class would make sense only if `SendMessage` is used by other actions that sends messages not related to Sentry (as an interface with `get_message` and `send_message` methods).

spidermon/contrib/actions/sentry/__init__.py
muzaffaryousaf: @rennerocha Even with available patterns, that's what already happening for different things. e.g. subject. The context that you have shared is only being used for something to include in report itself not for titles / subjects or something for external to the report.

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -0,0 +1,124 @@
+from __future__ import absolute_import
+from __future__ import print_function
```
rennerocha: Just noticed now. As we are not using `print` in this action, this import can be removed without problems.

spidermon/contrib/actions/sentry/sentry.py
muzaffaryousaf: Going to have single class now, since we don't need to split it by template and actions being made.

spidermon/contrib/actions/sentry/__init__.py
muzaffaryousaf: Already did :)

Issue #82, opened 2019-01-17
Add SPIDERMON_ENGINE_STOP_MONITORS


HassanQamar07:
I want to monitor if the file was uploaded to S3 or not. I've subclassed the FeedExporter and add a flag in stats when the file is uploaded. But the issue the `SPIDERMON_SPIDER_CLOSE_MONITOR` runs all the monitors even before the file is uploaded, this is because `FeedExporter` and spidrmon `SPIDERMON_SPIDER_CLOSE_MONITOR` both runs on the spider_close signal. 

To monitor this we need to run some monitors on the `engine_stopped` signal.

heylouiz:
@HassanQamar07 Can I implement this change?

rennerocha:
Merged.

PR #96, opened 2019-02-13
Add SPIDERMON_ENGINE_STOP_MONITORS setting to list monitors to be executed when Scrapy engine is stopped

HassanQamar07:
This implements the feature https://github.com/scrapinghub/spidermon/issues/82


rennerocha:
As @ejulio said, code looks fine, but it would be very helpful if you could include some unit tests.

raphapassini:
We are needing this feature in a project, I'll take care of the missing unit tests and documentation if you guys don't mind. @HassanQamar07 ?

raphapassini:
I did add the test, there's a little problem but is due to the fact that `unittest.mock` used in `test_extension.py` is not available on Python2. We can easily fix this by adding the `mock` library to the `requirements.txt` but I would like to hear comments about the test first, maybe you guys have ideas about how to solve without mock, or you have more ideas of tests.

tests/test_extension.py
```diff
@@ -0,0 +1,35 @@
...
+
+@pytest.fixture
+def crawler(spidermon_enabled_settings):
+    crawler = Crawler(Spider, settings=spidermon_enabled_settings)
+    crawler.spider = Spider('dummy')
+    return crawler
+
+
+def test_engine_stopped_suites_should_run(crawler):
+    "The suites defined at engine_stopped_suites should be loaded and run """
```
ejulio: Maybe I'm missing something, but there shouldn't be `""" some text """`

tests/test_extension.py
```diff
@@ -0,0 +1,35 @@
...
+    engine_stopped_suites = [
+        'tests.fixtures.suites.Suite01'
+    ]
+    spidermon = Spidermon(crawler, engine_stopped_suites=engine_stopped_suites)
+    spidermon.engine_stopped_suites[0].run = mock.MagicMock()
+    spidermon.engine_stopped()
+    assert spidermon.engine_stopped_suites[0].__class__.__name__ == 'Suite01'
+    assert spidermon.engine_stopped_suites[0].run.called
```
ejulio: I'm fine using `mock` and installing through `requirements`.
Here, probably I would assign `mock` to it's own variable to make the assertions shorter.
Also, probably would use `my_mock.assert_called_once()` or other assertion built-in with `mock`.

tests/test_extension.py
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+
+@pytest.fixture
+def crawler(spidermon_enabled_settings):
+    crawler = Crawler(Spider, settings=spidermon_enabled_settings)
+    crawler.spider = Spider('dummy')
+    return crawler
+
+
```
ejulio: Also, this test only tests that _when calling `engine_stopped`, it'll call the proper suites_.
I think that another important test would be to check if Spidermon subscribed properly to `engine_stopped` signal from Scrapy.

tests/test_extension.py
rennerocha: I believe that when black is executed, this will be fixed.

heylouiz:
@raphapassini @rennerocha @ejulio 
I made the requested changes, I had some problems to add `mock` as a requirement because I wanted to make sure it only installs when running with Python2, but I had problems with the tox configuration. I ended up putting it on a dedicated file for tests on requirements directory.

tests/test_extension.py
heylouiz: @ejulio I tried to use `assert_called_once()` but it is not supported in python 3.5. I went back from Raphael's approach.

requirements/tests.txt
```diff
@@ -0,0 +1 @@
+mock
```
rennerocha: As this is required only for Python 2.7, you can include an environment marker, so it won't be installed in newer Python versions (https://www.python.org/dev/peps/pep-0508/#environment-markers)

spidermon/contrib/scrapy/extensions.py
```diff
@@ -62,13 +67,16 @@ def from_crawler(cls, crawler):
...
+            engine_stopped_suites=crawler.settings.getlist('SPIDERMON_ENGINE_STOP_MONITORS'),
             spider_opened_expression_suites=crawler.settings.getlist('SPIDERMON_SPIDER_OPEN_EXPRESSION_MONITORS'),
             spider_closed_expression_suites=crawler.settings.getlist('SPIDERMON_SPIDER_CLOSE_EXPRESSION_MONITORS'),
+            engine_stopped_expression_suites=crawler.settings.getlist('SPIDERMON_ENGINE_STOP_EXPRESSION_MONITORS'),
             expressions_monitor_class=crawler.settings.get('SPIDERMON_EXPRESSIONS_MONITOR_CLASS'),
             periodic_suites=crawler.settings.getdict('SPIDERMON_PERIODIC_MONITORS'),
         )
         crawler.signals.connect(ext.spider_opened, signal=signals.spider_opened)
         crawler.signals.connect(ext.spider_closed, signal=signals.spider_closed)
+        crawler.signals.connect(ext.engine_stopped, signal=signals.engine_stopped)
```
rennerocha: This line could be tested like:
https://github.com/scrapinghub/spidermon/pull/120

tests/test_extension.py
```diff
@@ -0,0 +1,92 @@
...
+    return _crawler
+
+
+@pytest.fixture
+def suites():
+    return ['tests.fixtures.suites.Suite01']
+
+
+def test_spider_opened_suites_should_run(crawler, suites):
```
rennerocha: As `crawler` argument is a function, maybe a better name for the fixture would be `get_crawler`...

tests/test_extension.py
```diff
@@ -0,0 +1,92 @@
...
+    """The suites defined at spider_opened_suites should be loaded and run """
+    crawler = crawler()
```
rennerocha: ... then you could change to `crawler = get_crawler()` which looks more readable.

rennerocha:
@heylouiz 
You need to apply black to format the code.

tests/test_extension.py
ejulio: Are you sure?https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once

spidermon/contrib/scrapy/extensions.py
ejulio: Nothing against the approach, I just think that we only need to test if it _connected_ to the signal.
The test of being called when the signal is triggered, is already handled by scrapy.

heylouiz: But I think it's best to test this on your PR  right? or on this if yours get approved first.

heylouiz:
@ejulio 
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once
`New in version 3.6.`

It was failing on python 3.5 build on Travis.

tests/test_extension.py
heylouiz: @ejulio I replied on the wrong place, pasting here too:

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once
`New in version 3.6.`

It was failing on python 3.5 build on Travis.

ejulio: My bad!
But, there are alternatives like `assert_any_call` or `assert_has_calls(1)`

heylouiz: @ejulio I ended up using assert_called_once_with(mock.ANY), what do you think?

ejulio: Looks good to me.
Not sure what is better, if `assert_any_call` or `assert_called_once_with(mock.ANY)`.

Issue #72, opened 2019-01-09
Make basic monitors part of the spidermon.


muzaffaryousaf:
We have to add basic monitors for almost all the projects and the goal is to make such monitors part of the tool/spidermon itself and upon enabling them, the tool should send notifications automatically when integrated with a project. 

- [ ] Basic monitors inside the spidermon.
- [ ] One should be able to add customized monitors in a project in addition to these basic monitors.
- [ ] On should be able to disable these monitors and add their own monitors.

muzaffaryousaf:
@rennerocha @raphapassini FYI, I've added list of basic monitors and expectations here.

muzaffaryousaf:
Hi @raphapassini Would you be able to take this on ? Thanks.

muzaffaryousaf:
@raphapassini I'm assigning this to you as we spoke yesterday. Please let me know otherwise. Thanks.

raphapassini:
I'll start the implementation today, you should here from me soon :)

raphapassini:
The PR is finished and waiting for review!

rennerocha:
Merged

PR #97, opened 2019-02-14
Basic monitors

raphapassini:
This should close this [Issue #72](https://github.com/scrapinghub/spidermon/issues/72)


spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    default_codes = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
+
+    @monitors.name('Should not hit the limit of unwanted http status')
+    def test_check_unwanted_http_codes(self):
+        max_errors = self.get_settings(SPIDERMON_MAX_UNWANTED_HTTP_CODES, 1)
+        error_codes = self.get_settings(SPIDERMON_UNWANTED_HTTP_CODES,
+                                        self.default_codes)
+        for code in error_codes:
+            times = self.stats.get(
```
ejulio: `counts` or `status_counts` I think would be better.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,145 @@
...
+
+    You can also configured the **maximum** number of times that those
+    response codes with ``SPIDERMON_MAX_UNWANTED_HTTP_CODES`` the default
+    value is ``1``.
+    """
+    default_codes = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
+
+    @monitors.name('Should not hit the limit of unwanted http status')
+    def test_check_unwanted_http_codes(self):
+        max_errors = self.get_settings(SPIDERMON_MAX_UNWANTED_HTTP_CODES, 1)
```
ejulio: What do you think of a threshold for each error code?
Probably, I accept up to 10 404s, but up to 50 503s.

ejulio: Also, not sure if `max_errors` or `max_status_count`.
An HTTP status code is not necessarily an error.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,145 @@
...
+                                        self.default_codes)
+        for code in error_codes:
+            times = self.stats.get(
+                'downloader/response_status_count/{}'.format(code), 0)
+            msg = 'Found {} Responses with status code={} - '\
+                  'This exceed the limit of {}'.format(times, code, max_errors)
+            self.assertTrue(times < max_errors, msg=msg)
+
+
+class SpiderCloseMonitorSuite(MonitorSuite):
```
ejulio: Do we need to provide a default suite or only the basic monitors?
I ask because of the CLI client that, probably, will create this suite.

tests/test_monitors.py
```diff
@@ -0,0 +1,184 @@
...
+
+def new_crawler():
+    return Crawler(Spider)
+
+
+def new_suite(monitors):
+    return MonitorSuite(monitors=monitors)
+
+
+def test_needs_to_configure_item_count_monitor():
```
ejulio: What do you think of wrapping these tests in a `TestCase`?
This would make it possible add a `setUp` method to instantiate the crawler, spider and all other studd that is repeated among some of the tests.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,145 @@
...
+                   SPIDERMON_MAX_ERROR = {
+                        'my_spider': 100,
+                        'other_spider': 10,
+                   }
+
+        """
+        settings_value = self.crawler.settings.get(name.upper())
+        if type(settings_value) == dict:
+            settings_value = settings_value.get(self.spider.name, None)
+        spider_value = getattr(self.spider, name.lower(), None)
```
ejulio: I'm not sure if this a standard practice within scrapy, to default settings to spider attributes.
Mostly because we could achieve the same by just keeping the settings, but using the `custom_settings` _dict_ in the spider.

ejulio: But, if we keep this approach, I think we should document it somewhere.

raphapassini: It's documented in the doc string of the function. I can easily change that.
Let's wait for more comments on this one.

spidermon/contrib/scrapy/monitors.py
raphapassini: We need this monitor for sure, if this is the question here.
Establishing arbitrary values about what I think should be the threshold is a bad idea, it should be flexible enough so you can configure the way you want in your project.

spidermon/contrib/scrapy/monitors.py
raphapassini: This can be discussed further, but I think it's a **nice** approach to have more 5 lines her, this way the developer doesn't need to write this line in his own code nor the CLI needs to generate it :)

spidermon/contrib/scrapy/monitors.py
raphapassini: You're talking about having something like:

``python
SPIDERMON_UNWANTED_HTTP_CODES = {
    '500: 50,
    '400': 10,
}``

I like the flexibility of the approach but we actually need it? You can't have a threshold to rule then all?
Would like to hear more comments on this one as well.

victor-torres: It would be good but we can live without it for a while.

victor-torres: It would be good but we can live without it for a while.

tests/test_monitors.py
rennerocha: We use pytest, so it is not necessary to create a `TestCase`. 
As all tests creates a crawler and a spider, you could use [setup_function](https://docs.pytest.org/en/latest/xunit_setup.html?highlight=setup_function) function that will be executed before each function.

tests/test_monitors.py
```diff
@@ -0,0 +1,184 @@
...
+            'crawler': crawler,
+            'spider': spider,
+            'job': None,
+    }
+    runner.run(suite, **data)
+    for r in runner.result.monitor_results:
+        assert('NotConfigured' in r.error)
+
+
+def test_item_count_monitor_should_fail():
```
rennerocha: I think the test will be more robust, if you create more scenarios. You can use [pytest.fixtures](https://docs.pytest.org/en/latest/example/parametrize.html?highlight=pytest.fixture#indirect-parametrization-with-multiple-fixtures) to achieve it.

Some scenarios I can imagine for this test:
- number of items equal to SPIDERMON_MIN_ITEMS
- number of items greater than SPIDERMON_MIN_ITEMS
- number of items lower than SPIDERMON_MIN_ITEMS
- SPIDERMON_MIN_ITEMS not a valid integer

Rename the test would be good, if you use the fixtures.

tests/test_monitors.py
```diff
@@ -0,0 +1,184 @@
...
+            'crawler': crawler,
+            'spider': spider,
+            'job': None,
+    }
+    runner.run(suite, **data)
+    for r in runner.result.monitor_results:
+        assert('Extracted 10 items, the expected minimum is 100' in r.error)
+
+
+def test_item_count_monitor_should_pass():
```
rennerocha: This test can be merged with the latest test, using a pytest.fixture.

tests/test_monitors.py
```diff
@@ -0,0 +1,184 @@
...
+            'stats': crawler.stats.get_stats(),
+            'crawler': crawler,
+            'spider': spider,
+            'job': None,
+    }
+    runner.run(suite, **data)
+    assert(runner.result.monitor_results[0].error is None)
+
+
+def test_finished_reason_monitor_should_fail():
```
rennerocha: This also could be done with pytest.fixtures.

spidermon/contrib/scrapy/monitors.py
rennerocha: Not sure about that, but if we do, we will need more info in the docs explaining how to enable it.

raphapassini: There's documentation about how to enable this Suite, you think the documentation isn't clear or you didn't found it @rennerocha ?

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,145 @@
...
+
+SPIDERMON_MIN_ITEMS = 'SPIDERMON_MIN_ITEMS'
+SPIDERMON_MAX_ERROR = 'SPIDERMON_MAX_ITEMS'
+SPIDERMON_EXPECTED_FINISH_REASONS = 'SPIDERMON_EXPECTED_FINISH_REASONS'
+SPIDERMON_MAX_UNWANTED_HTTP_CODES = 'SPIDERMON_MAX_UNWANTED_HTTP_CODES'
+SPIDERMON_UNWANTED_HTTP_CODES = 'SPIDERMON_UNWANTED_HTTP_CODES'
+
+
+class BaseScrapyMonitor(Monitor, SpiderMonitorMixin):
+    longMessage = False
```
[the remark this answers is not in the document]
raphapassini: Yes it's declared here: https://docs.python.org/3/library/unittest.html#unittest.TestCase.debug

spidermon/contrib/scrapy/monitors.py
ejulio: @rennerocha and @raphapassini  This was my idea, sorry if I wasn't clear enough :)
For me, the default threshold for all error codes are fine, just wrote a thought that could be helpful to make it more flexible.
Agree with @victor-torres and @rennerocha  that we can live with a single value for all codes for now, and in the future we could change it.

spidermon/contrib/scrapy/monitors.py
ejulio: @raphapassini, I got the docstring, but I was talking about the spidermon docs.
Because this is a feature the API user would need to know without going to the source code.

docs/source/monitors.rst
```diff
@@ -130,6 +130,23 @@ Here is an example of how to configure a new monitor suite in your project:
 
       ``result`` stats of the spider execution
 
+The Basic Monitors
```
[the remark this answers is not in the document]
victor-torres: I think it's a nice idea!

spidermon/contrib/scrapy/monitors.py
rennerocha: @raphapassini I found the docs now :-)

spidermon/contrib/scrapy/monitors.py
raphapassini: I'm changing this @ejulio but I'm using autodoc, so the docstring _from_ the function *is* going to the documentation :)

spidermon/contrib/scrapy/monitors.py
raphapassini: I like the suggestion it was implemented :)

tcurvelo:
I'm just came for the caipirinhas :tropical_drink: :brazil:

docs/source/monitors.rst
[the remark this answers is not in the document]
raphapassini: Which tutorial you guys are talking about the ``Getting Started`` section?

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,146 @@
+from spidermon import monitors, MonitorSuite, Monitor
+from spidermon.exceptions import NotConfigured
+from ..monitors.mixins.spider import SpiderMonitorMixin
+
+SPIDERMON_MIN_ITEMS = 'SPIDERMON_MIN_ITEMS'
+SPIDERMON_MAX_ERROR = 'SPIDERMON_MAX_ERROR'
```
victor-torres: What about _SPIDERMON_MAX_ERROR**S**_ (plural)?

(MySQL uses _max_error_count_, but we could stick with the shorter version)

docs/source/monitors.rst
```diff
@@ -130,6 +130,23 @@ Here is an example of how to configure a new monitor suite in your project:
...
+
+Spidermon have some batteries included :)
+
+.. automodule:: spidermon.contrib.scrapy.monitors
+    :members: FinishReasonMonitor, ItemCountMonitor, LogMonitor, UnwantedHTTPCodesMonitor,
+
+Is there a **Basic Scrapy Suite** ready to use?
+------------------------------------------------
+
+Of course there is! We really want to make it easy for you to monitor your spiders ;)
```
victor-torres: You're still missing the comma after the 'Of course' expression. 

> Of course, there is! We really want to make it easy for you to monitor your spiders ;)

spidermon/contrib/scrapy/monitors.py
```diff
@@ -101,12 +101,12 @@ class UnwantedHTTPCodesMonitor(BaseScrapyMonitor):
             code: 10
             for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]}
 
-    **WARNING**: You can have this settings by spider like::
+    **WARNING**: You CAN NOT have this settings by spider like::
```
victor-torres: Both cannot and can not are acceptable spellings, but the first is much more usual. You would use can not when the ‘not’ forms part of another construction such as ‘not only’. For example:

> These green industries can not only create more jobs, but also promote sustainable development of the land.

Reference: https://en.oxforddictionaries.com/usage/cannot-or-can-not

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,146 @@
...
+               .. code-block:: python
+
+                   SPIDERMON_MAX_ERRORS = {
+                        'my_spider': 100,
+                        'other_spider': 10,
+                   }
+
+        """
+        settings_value = self.crawler.settings.get(name.upper())
+        if type(settings_value) == dict:
```
rennerocha: I am not 100% certain, but I believe `type(settings_value) is dict` is a better way to check if this variable is a dict.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,146 @@
...
+    def test_minimum_number_of_items(self):
+        item_extracted = getattr(self.stats, 'item_scraped_count', 0)
+        msg = 'Extracted {} items, the expected minimum is {}'.format(
+            item_extracted, self.minimum_threshold)
+        self.assertTrue(
+            item_extracted >= self.minimum_threshold, msg=msg
+        )
+
+
+@monitors.name('Log Monitor')
```
rennerocha: This is not a "Log monitor". I believe a better name would be "Number of error monitor"

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,146 @@
...
+SPIDERMON_MIN_ITEMS = 'SPIDERMON_MIN_ITEMS'
+SPIDERMON_MAX_ERRORS = 'SPIDERMON_MAX_ERRORS'
+SPIDERMON_EXPECTED_FINISH_REASONS = 'SPIDERMON_EXPECTED_FINISH_REASONS'
+SPIDERMON_UNWANTED_HTTP_CODES = 'SPIDERMON_UNWANTED_HTTP_CODES'
+
+
+class BaseScrapyMonitor(Monitor, SpiderMonitorMixin):
+    longMessage = False
+
+    def get_settings(self, name, default=None):
```
rennerocha: I still think that using `custom_settings` to define particular values for each spider is more consistent with the way Scrapy uses for defining settings for each spider, but it is fine as it is.

rennerocha: There is no tests for this method.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,146 @@
...
+@monitors.name('Finish Reason Monitor')
+class FinishReasonMonitor(BaseScrapyMonitor):
+    """Check if a job has a expected finish reason.
+    You can configure the expected reason with the
+    ``SPIDERMON_EXPECTED_FINISH_REASONS``, it should be an ``iterable`` of
+    valid finish reasons.
+
+    The default value of this settings is: ``['finished', ]``."""
+    @monitors.name('Should have the expected finished reason(s)')
+    def test_should_not_have_errors(self):
```
rennerocha: This test name doesn't match with the monitor purpose.

spidermon/contrib/scrapy/monitors.py
ejulio: I don't know if they are required either.
This method is not part of the public API, so I don't think we need to test it directly.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+@monitors.name('Extracted Items Monitor')
+class ItemCountMonitor(BaseScrapyMonitor):
+    """Check if spider extracted the minimum number of items.
+    You can configure it using ``SPIDERMON_MIN_ITEMS`` setting.
+    There's **NO** default value for this setting, if you try to use this
+    monitor without setting it, it'll raise a ``NotConfigured`` exception.
+    """
+    def run(self, result):
```
ejulio: @raphapassini, any specific reason to use `run` instead of `__init__`?
Also, `run` is executed when the spider is started or at the end of the job?

spidermon/contrib/scrapy/monitors.py
ejulio: Another option is `ErrorCountMonitor`

spidermon/contrib/scrapy/monitors.py
victor-torres:
```
isinstance(...)
    isinstance(object, class-or-type-or-tuple) -> bool
```

spidermon/contrib/scrapy/monitors.py
raphapassini: It's tested because the other monitors relies on it, so it's called and tested. Please refer to the coverage.

spidermon/contrib/scrapy/monitors.py
raphapassini: When the ``__init__`` is called you don't have the ``data`` information where you can get the ``crawler``, ``settings``, ``spider`` etc. It __doesn't__ work in the ``__init__`` method.

PR #92, opened 2019-02-11
Any color you like

manycoding:
#91 
I didn't exclude any files, but it looks right. I kind of want to separate **requirements.txt** from the **tox.ini** too and use **setup.py** `extras`, but I am not sure if it belongs to this pr.

Last but not least, `pep8` section in tox doesn't really need all requirements, but I am not sure how to avoid default ones in `testenv`.

Issue #91, opened 2019-02-11
Black to handle pep8 and all code formating


manycoding:
I have noticed here and there formatting issues in prs, I suggest looking at https://github.com/ambv/black.

The setup is simple - add few lines of settings to `tox.ini`, autoformat the code with `black ./`, and add a line to CI to check everything is ok.

The main benefit - it allows to focus on code, it brings consistency, integrations with IDEs are good.
The main cons for me - it's a pre-release package, and why it's very stable `-pre` is a special case.

I am ready to help.

manycoding:
@rennerocha @raphapassini

rennerocha:
I like the idea of using something like black to format our code. I have been using it in some projects and it removes friction between developers coding styles.

PR #99, opened 2019-02-18
Any colour you like

manycoding:
#91 
As per https://github.com/scrapinghub/spidermon/pull/92#issuecomment-464754628
This commit adds Black and CI setup

@raphapassini


raphapassini:
+1 to merge

rennerocha:
I understand what we want to accomplish but I didn't understand what this PR will do.

The build will fail if the following command lists files that needs to be changed? 
`black ./ --check --diff`

manycoding:
@rennerocha Yes. That's a good way to ensure the uniform style. I preformatted the files in another pr, so it won't fail after merge.
Basically `black  ./ --check --diff` is the same as `flake .`

rennerocha:
@manycoding thanks for the explanation :-)

Issue #89, opened 2019-02-07
Are optional dependencies really optional?


wRAR:
This is probably related to #88.

It seems to me that if you want to use spidermon with Scrapy you need to use `spidermon.contrib.scrapy.extensions.Spidermon`. That requires at least `jsonschema` (imported via `spidermon.python.factory`) from the optional `validation` feature. On the other hand, the contents of the optional `monitoring` feature look important for any spidermon usage. Are there real use cases where one or both optional features can be skipped?

Disclaimer: I'm not familiar with optional installation features that much, as most modules I've used don't use them.

rennerocha:
@wRAR please see my comment in #88 . IMHO, optional requirements are optional, as the `core` of spidermon functionality still work without them .

andrewbaxter:
FWIW I'm getting an exception since it seems like jsonschema gets imported when using the extension:

```
Traceback (most recent call last):
  File "/myproject/env/bin/scrapy", line 11, in <module>
    sys.exit(execute())
  File "/myproject/env/lib/python3.6/site-packages/scrapy/cmdline.py", line 150, in execute
    _run_print_help(parser, _run_command, cmd, args, opts)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/cmdline.py", line 90, in _run_print_help
    func(*a, **kw)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/cmdline.py", line 157, in _run_command
    cmd.run(args, opts)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/commands/crawl.py", line 57, in run
    self.crawler_process.crawl(spname, **opts.spargs)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/crawler.py", line 170, in crawl
    crawler = self.create_crawler(crawler_or_spidercls)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/crawler.py", line 198, in create_crawler
    return self._create_crawler(crawler_or_spidercls)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/crawler.py", line 203, in _create_crawler
    return Crawler(spidercls, self.settings)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/crawler.py", line 55, in __init__
    self.extensions = ExtensionManager.from_crawler(self)
  File "/myproject/env/lib/python3.6/site-packages/scrapy/middleware.py", line 58, in from_crawler
[...]

PR #100, opened 2019-02-19
Included jsonschema as required dependency

rennerocha:
Used by PythonExpressionsMonitor (https://github.com/scrapinghub/spidermon/blob/master/spidermon/python/factory.py#L34)


rennerocha:
Fixes #89

PR #112, opened 2019-02-26
Fix problem caused by backward incompatible change in jsonschema 3.0.0

rennerocha:
After release of jsonschema 3.0.0, it is possible to validate data against draft6 and draft7. The keywords _exclusiveMinimum_ and _exclusiveMaximum_ changed in a backwards incompatible change which required an update in the existing tests.


rennerocha:
We will have 3 classes instead of one if we follow this suggestion. As we don't have a big number of test scenarios I believe that maintaining just one class is enough and clear.

PR #119, opened 2019-03-05
Get schema from url

manycoding:
A first attempt. Presumably, it works.
I haven't figured out the tests yet, this whole TestCase structure makes mi dizzy.

The questions is how to properly mock `get_contents`


spidermon/utils/web.py
```diff
@@ -0,0 +1,17 @@
+from urllib.parse import urlparse
+from urllib.request import urlopen
+
+
+def is_url(path: str) -> bool:
```
rennerocha: Function annotations doesn't work in Python 2.7.

rennerocha: We already have a function to verify if a string is a valid URL:
`spidermon.contrib.validation.utils.is_valid_url`

spidermon/contrib/scrapy/pipelines.py
```diff
@@ -84,7 +85,9 @@ def set_validators(loader, schema):
     @classmethod
     def _load_jsonschema_validator(cls, schema):
         if isinstance(schema, six.string_types):
-            if schema.endswith(".json"):
+            if web.is_url(schema):
+                schema = json.loads(web.get_contents(schema))
```
rennerocha: What if the content returned is not a valid JSON?

tests/contrib/scrapy/test_pipelines.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
-from unittest import TestCase
+from unittest import mock, TestCase
```
rennerocha: `unittest.mock` is not available in Python 2.7
You need to install `mock` library for this Python version and import accordingly.

spidermon/contrib/scrapy/pipelines.py
manycoding: You will get an exception. How it's supposed to work?
Probably I need to put .json back, at least it's dummy proof.

spidermon/utils/web.py
manycoding: `is_valid_url` doesn't allow s3 paths like s3://bucket/file and it has a different purpose - validate with schema.

tests/contrib/scrapy/test_pipelines.py
```diff
@@ -210,6 +212,25 @@ class PipelineJSONSchemaValidator(PipelineTest):
     ]
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 4), reason="mock requires python3.4 or higher"
+)
+def test_validator_from_url():
```
manycoding: I haven't figured out how to create mock properly with the current DataTest structure so I added a simple pytest style test.

spidermon/contrib/scrapy/pipelines.py
raphapassini: You can be friendly to the user with something like:
```python
try:
     schema = json.loads(web.get_contents(schema))
except json.JSONDecodeError:
     self.logger.exception(f'Could not parse json schema downloaded from {schema}')
```

rennerocha: I agree with @raphapassini . In this case it would be better to provide a better error message to the user being more explicit over the problem that occurred.

spidermon/utils/web.py
```diff
@@ -0,0 +1,23 @@
...
+    from urllib import urlopen
+
+
+def is_schema_url(path):
+    result = urlparse(path)
+    try:
+        if all([result.scheme, result.netloc, result.path]):
+            return True
+        return False
+    except AttributeError:
```
rennerocha: The only call that could raise `AttributeError` is `urlparse` that is outside the `try` block.

spidermon/utils/web.py
```diff
@@ -0,0 +1,23 @@
...
+        return False
+
+
+def get_contents(url):
+    with urlopen(url) as f:
```
rennerocha: Same as https://github.com/scrapinghub/spidermon/pull/119/files#r263525248, you could be more friendly to the user handling `URLError` and logging the reason for the failure.

tests/contrib/scrapy/test_pipelines.py
```diff
@@ -210,6 +212,25 @@ class PipelineJSONSchemaValidator(PipelineTest):
...
+    from unittest import mock
+
+    with mock.patch("spidermon.utils.web.get_contents") as mocked_get_contents:
+        mocked_get_contents.return_value = test_schema_string
+        settings = {SETTING_SCHEMAS: {TestItem: "https://fixtures.com/testschema.json"}}
+        test_item = TestItem()
+        crawler = get_crawler(settings_dict=settings)
+        pipe = ItemValidationPipeline.from_crawler(crawler)
```
rennerocha: Maybe checking the stats of the pipeline isn't really testing if the JSON validator is using the content returned by the schema you provided. I thought in two different tests that may check better this scenario:
- Verify if the `ItemValidationPipeline` returned contains inside `validators` a `JSONSchemaValidator` containing `test_schema_string` in `_schema` or;
- When creating an `ItemValidationPipeline`, verify if `JSONSchemaValidator` was created using `test_schema_string` and this validator appears in `validators`of the pipeline.

tests/contrib/scrapy/test_pipelines.py
```diff
@@ -210,6 +212,25 @@ class PipelineJSONSchemaValidator(PipelineTest):
     ]
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 4), reason="mock requires python3.4 or higher"
+)
+def test_validator_from_url():
+    from unittest import mock
+
+    with mock.patch("spidermon.utils.web.get_contents") as mocked_get_contents:
```
rennerocha: You have suggested me in other PR, so why you don't use `pytest-mock` and avoid the importing here?

tests/utils/test_web.py
```diff
@@ -0,0 +1,16 @@
+import pytest
```
rennerocha: Missing tests for `get_contents`.

spidermon/contrib/scrapy/pipelines.py
manycoding: How do I access logger?

spidermon/utils/web.py
manycoding: Not really, if schema doesn't have scheme result.scheme will raise an attribute error.

tests/contrib/validation/jsonschema/test_tools.py
```diff
@@ -0,0 +1,41 @@
+import os
+import sys
+
+import pytest
+from spidermon.contrib.validation.jsonschema.tools import get_schema_from
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 5), reason="we don't want to test json on different versions"
```
manycoding: JSON seem to return different errors for Python 2 and 3, so considering we will abandon it support soon I prudently decided to skip it.

manycoding:
@raphapassini @rennerocha Please review, I added tests and addressed your comments.

raphapassini:
LGTM

spidermon/utils/web.py
```diff
@@ -0,0 +1,20 @@
+import sys
+import logging
+from six.moves.urllib.parse import urlparse
```
rennerocha: The functions of this module are so small, and used only in spidermon/contrib/validation/jsonschema/tools.py that I think they would be better placed in the same module. This would improve readability, as we won't need to jump between modules to see functions simple like that.

tests/contrib/scrapy/test_pipelines.py
```diff
@@ -69,7 +71,7 @@ class PipelineTest(six.with_metaclass(PipelineTestCaseMetaclass, TestCase)):
 
 
 class DataTest(object):
-    def __init__(self, name, item, cases, settings={}):
+    def __init__(self, name, item, cases, settings=dict()):
```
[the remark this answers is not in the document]
manycoding: That's a common Python pitfall https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

spidermon/utils/web.py
```diff
@@ -0,0 +1,20 @@
...
+    except AttributeError:
+        return False
+
+
+def get_contents(url):
+    try:
+        with urlopen(url) as f:
+            return f.read().decode("utf-8")
+    except Exception as e:
+        logging.exception(str(e) + "\nFailed to get '{}'".format(url))
```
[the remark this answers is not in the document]
manycoding: `scheme` will be empty, but netloc or path could be None
```
>>> all([urlparse("anything").scheme, urlparse("anything").netloc, urlparse("anything").scheme.path])
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'path'
```

manycoding:
@rennerocha done 😄

rennerocha:
@manycoding I will review it today ...

PR #120, opened 2019-03-07
Test if spidermon is connected to desired Scrapy signals

rennerocha:
The included tests verifies that Spidermon will handle spider_opened and spider_closed Scrapy signals.


tests/test_spidermon_signal_connect.py
```diff
@@ -0,0 +1,47 @@
...
+    with mock.patch.object(Spidermon, "spider_opened") as spider_opened_method:
+        crawler = get_crawler(
+            settings_dict={
+                "SPIDERMON_ENABLED": True,
+                "EXTENSIONS": {"spidermon.contrib.scrapy.extensions.Spidermon": 100},
+            }
+        )
+        spider = Spider.from_crawler(crawler, "example.com")
+        crawler.signals.send_catch_log(signal=signals.spider_opened, spider=spider)
+        assert spider_opened_method.called, "spider_opened not called"
```
manycoding: Why not `spider_opened_method.assert_called_once_with(params?)`

tests/test_spidermon_signal_connect.py
```diff
@@ -0,0 +1,47 @@
...
+def get_crawler(settings_dict=None):
+    runner = CrawlerRunner(settings_dict)
+    return runner.create_crawler(Spider)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 4), reason="requires python3.4 or higher")
+def test_spider_opened_connect_signal():
+    from unittest import mock
+
+    with mock.patch.object(Spidermon, "spider_opened") as spider_opened_method:
```
manycoding: If you use pytest-mock you can simplify it as (and avoid import):

    spider_opened_method = mocker.patch.object(Spidermon, "spider_opened")

another example https://github.com/scrapinghub/gatf/blob/2e7208228d8daebdb23e84c88483f16b2cbfb487/tests/test_gatf.py#L50

Also I always use `autospec`, since without it you can use imaginable methods and not even know the test doesn't work

manycoding: I always use `autospec`, since without it you can use imaginable methods and not even know the test doesn't work

rennerocha: `autospec` didn't work . Do you have an idea why?

tests/test_spidermon_signal_connect.py
```diff
@@ -14,34 +14,35 @@ def get_crawler(settings_dict=None):
...
-        crawler = get_crawler(
-            settings_dict={
-                "SPIDERMON_ENABLED": True,
-                "EXTENSIONS": {"spidermon.contrib.scrapy.extensions.Spidermon": 100},
-            }
-        )
-        spider = Spider.from_crawler(crawler, "example.com")
-        crawler.signals.send_catch_log(signal=signals.spider_opened, spider=spider)
-        assert spider_opened_method.called, "spider_opened not called"
+    spider_opened_method = mocker.patch.object(Spidermon, "spider_opened")
```
manycoding: `autospec=True` will do the trick.
Don't forget to specify it everywhere or mock.asser_called_with_nonce and such funny staff will pass.

manycoding:
@rennerocha I stand for always using `autospec=True`, any reason why you didn't?
Here's some explanation mocks.https://medium.com/ryans-dev-notes/python-autospeccing-72c2a5ba5e28

Issue #102, opened 2019-02-19
Use logging library in spidermon


rennerocha:
Some actions uses `print` for logging:
- https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/email/__init__.py#L77
- https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/slack/__init__.py#L179

Python built-in [logging](https://docs.python.org/3.5/library/logging.html) library should be used instead (like 
[here](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/sentry/__init__.py#L3)).

leovictorsr:
@ejulio hello Julio, we spoke via email, I'll work on this issue as my first for Spidermon.

ejulio:
hey @leovictorsr.
That's awesome to hear!
If you need anything, just ping me I'll try to help you :smile:

PR #121, opened 2019-03-08
Replace print on actions with logging library

leovictorsr:
This implements the feature #102

### DIFF
--- a/spidermon/contrib/actions/email/__init__.py
+++ b/spidermon/contrib/actions/email/__init__.py
@@ -1,7 +1,9 @@
 from __future__ import absolute_import
-from __future__ import print_function
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
+import logging
+
+logger = logging.getLogger(__name__)
 
 from spidermon.contrib.actions.templates import ActionWithTemplates
 from spidermon.exceptions import NotConfigured
@@ -86,12 +88,10 @@ class SendEmail(ActionWithTemplates):
 
     def run_action(self):
         message = self.get_message()
-        if not self.fake:
-            self.send_message(message)
+        if self.fake:
+            logger.info(message.as_string())
         else:
-            print("-" * 40)
-            print(message.as_string())
-            print("-" * 40)
+            self.send_message(message)
 
     def get_subject(self):
         if self.subject:
--- a/spidermon/contrib/actions/slack/__init__.py
+++ b/spidermon/contrib/actions/slack/__init__.py
@@ -1,8 +1,10 @@
 from __future__ import absolute_import
-from __future__ import print_function
 import ast
 import json
+import logging
 from slackclient import SlackClient
+
+logger = logging.getLogger(__name__)
 
 from spidermon.contrib.actions.templates import ActionWithTemplates
 from spidermon.exceptions import NotConfigured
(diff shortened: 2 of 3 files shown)


spidermon/contrib/actions/email/__init__.py
```diff
@@ -89,9 +89,7 @@ def run_action(self):
         if not self.fake:
```
ejulio: Maybe you could improve this if statement as well.
It is easier to read affirmative sentences and negative ones.

spidermon/contrib/actions/slack/__init__.py
```diff
@@ -203,9 +203,9 @@ def run_action(self):
                 to=self.recipients, text=message, attachments=attachments
             )
         else:
-            print("message:", message)
-            print("attachments:", attachments)
-
+            logging.info(message)
+            logging.info(attachments)
+
```
ejulio: Just remove these whitespaces :smile:

manycoding:
>@rennerocha do we use logging directly or logger = logging.getLogger(__name__)?
@ejulio I think it depends if we want to have a file name in the log. Do we?

ejulio:
Agreed @manycoding.
Just ensuring conventions if there were any :smile:

spidermon/contrib/actions/slack/__init__.py
leovictorsr: Whitespace removed.

spidermon/contrib/actions/email/__init__.py
leovictorsr: Replace with affirmative condition.

rennerocha:
> > @rennerocha do we use logging directly or logger = logging.getLogger(**name**)?
> 
> @ejulio I think it depends if we want to have a file name in the log. Do we?

I believe it will be better to have it, or else it will be more difficult to identify where the log was generated.

@leovictorsr instead of using logging directly, include `logger = logging.getLogger(__name__)` in the beginning of the files (after the imports), to organise the loggers on a per-module basis. Then use `logger.info(your_variable)` when needed.

spidermon/contrib/actions/email/__init__.py
```diff
@@ -86,12 +86,10 @@ def from_crawler_kwargs(cls, crawler):
...
         message = self.get_message()
-        if not self.fake:
-            self.send_message(message)
+        if self.fake:
+            logging.info(message)
         else:
-            print("-" * 40)
-            print(message.as_string())
-            print("-" * 40)
+            self.send_message(message)
```
rennerocha: `message` is a `email.mime.multipart.MIMEMultipart` object, not a string.
Use `message.as_string()` instead of just `message` in the logs.

spidermon/contrib/actions/slack/__init__.py
```diff
@@ -203,8 +203,8 @@ def run_action(self):
                 to=self.recipients, text=message, attachments=attachments
             )
         else:
-            print("message:", message)
-            print("attachments:", attachments)
+            logging.info(message)
```
rennerocha: Use `logger.info(message)` and include `logger = logging.getLogger(__name__)` in the beginning of the module, to make the logger module-based (do the same in any other place we have logs like this).

spidermon/contrib/actions/email/__init__.py
leovictorsr: Done!

spidermon/contrib/actions/slack/__init__.py
leovictorsr: Done!
Please let me know if the way I did conforms with the guidelines because I didn't find any example on PEP08 or Python logging guidelines.

leovictorsr:
@rennerocha what do you think about logging unit tests? I think it would be good to have it and also to ensure all loggers use the same configuration in all modules.

spidermon/contrib/actions/email/__init__.py
```diff
@@ -86,12 +88,10 @@ def from_crawler_kwargs(cls, crawler):
 
     def run_action(self):
         message = self.get_message()
-        if not self.fake:
-            self.send_message(message)
+        if self.fake:
+            logging.info(message.as_string())
```
rennerocha: You miss the logger here.

leovictorsr: Done!

PR #127, opened 2019-03-14
Created CONTRIBUTING guide

rennerocha:


manycoding:
Isn't it too complex?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Types of Contributions
+----------------------
+
+Report Bugs
+~~~~~~~~~~~
+
+Report bugs at https://github.com/scrapinghub/spidermon/issues.
+
+If you are reporting a bug, please include:
```
manycoding: How about making an issue template and removing the list from here?

rennerocha: I like the idea of issues templates, but I also want to have some guidance for new contributors as soon as possible. I use other scrapinghub projects as reference for this file, so what do you think creating this first version and then we discuss issue template in a different PR?

muzaffaryousaf:
LGTM 👍

PR #129, opened 2019-03-14
Pipenv

manycoding:
I propose to use [pipenv](https://pipenv.readthedocs.io/en/latest/) as a development tool since it's simpler to manage with ~and use setup.py dependencies in tests~
Please vote if you like it or not with 👍 or 👎 

- [ ] Update docs


Gallaecio:
If we switch to setup.py for tests, I think we should have neither requirements.txt nor Pipfile on the repository.

Since this is a library meant to work with multiple versions of its dependencies, and not an application, I don’t think we need to enforce or even encourage a specific virtual environment tool on contributors.

manycoding:
@Gallaecio tox already does the tests using dependencies from `setup.py`, so it's out of question

Personally I find `pipenv` easier to manage than `pip`. Using requirements.txt enforces pip. As for different dependency versions, there's `tox` to help with that.

rennerocha:
What the problem are you trying to solve replacing the simple `requirements` file, that is known and used by the majority of Python developers with another tool? Does it justify to force every new contributor to learn a new tool to contribute to our project?

I see more drawbacks than gains forcing contributors to use a specific virtual environment tool, as it is more likely to find developers that don't know how to use `pipenv` (but knows how to use pip) than developers that may prefer `pipenv` (as a personal choice) but can use pip without problems.

manycoding:
@rennerocha I got your point. 
For me it's just:
1. simpler to have dev\core dependencies listed in a single file.
2. Checking the versions - pipfile generates pipfile.lock with all dependencies pinned. It's easier to see all those dependecies there than with pip. It's easier to find what's broken there.
Considering this, I don't really see big gains from using pipenv. From my side, I just like to propose good tools (I don't see why how it's bad to make developers to learn a new tool)

I just least the main benefits taken from https://pipenv.readthedocs.io/en/latest/:

- You no longer need to use pip and virtualenv separately. They work together.
- Managing a requirements.txt file can be problematic, so Pipenv uses Pipfile and Pipfile.lock to separate abstract dependency declarations from the last tested combination.
- Hashes are used everywhere, always. Security. Automatically expose security vulnerabilities.
- Strongly encourage the use of the latest versions of dependencies to minimize security risks arising from outdated components.
- Give you insight into your dependency graph (e.g. $ pipenv graph).
- Streamline development workflow by loading .env files.

Gallaecio:
Pipenv can already be used with a `requirements.txt` file; I use it, in fact. pip + venv cannot be used with a `Pipenv` file.

Gallaecio:
Pipenv can already be used with a `requirements.txt` file; I use it, in fact. pip + venv cannot be used with a `Pipfile` file.

Also, having `setup.py`, using `pipenv install -e .` or `pip install -e .` work as well, no need to use either `requirements.txt` or `Pipfile`.

alexander-matsievsky:
@manycoding I'd vote no on this. Pipenv seems to have been endorsed by PyPA [prematurely](https://www.google.com/search?hl=en&q=pipenv%20controversy) while still having [lots of major issues](https://github.com/pypa/pipenv/issues?q=is%3Aissue+is%3Aopen+sort%3Acomments-desc). It's better to hold on the tool till it stabilizes.

rennerocha:
Considering the comments from @Gallaecio and @alexander-matsievsky , I don't see any real gain in the development process if we start to use pipenv. We can discuss this later, but for the moment I will close this issue.

Issue #122, opened 2019-03-14
Create a base monitor to compare with previous crawls


ejulio:
Often, we compare the current crawling result with a previous one.
The idea here is to create some kind of base monitor that would make it a bit easier.
Basically, we need:
* Some sort of storage provider (we can use `python-scrapinghub` by default, but we need to make extensible for any kind of storage)
* A base monitor that would load last crawl data, an entry point to compare with the current one, storage the current crawl data.

Anything else you think is worth adding here?

ejulio:
I'll put some effort to move this issue :)
Here is what @rennerocha , @raphapassini and I discussed about the desired behavior to be added.

We'll provide two storage backends by default:

* `python-scrapinghub` to use past jobs stats in scrapy cloud.
* `local storae` to save the data to a /temp folder locally and then the user can persist it to wherever he wants
* In future version we can add support for AWS, GCS, and others

The backends will follow a basic interface with a method to load the past jobs stats and to store the current one.

We'll add a new setting so that the user can configure how many past jobs to keep.
`SPIDERMON_SOMETHING = 10`
Then, when running a monitor to compare with past jobs, it'll receive a list with stats from the last 10 jobs. Note that there might be limitations depending on the backend.

For monitoring, we'll add a new base monitor that will call the test methods with historical stats.
The idea is something like
```
class MyHistoricalMonitor(spidermon.HistoricalMonitor): # must work on the names :b
    @monitor('not far from the historical mean')
    def test_number_of_items_similar_to_historical_mean(self, past_stats):
        mean = compute_mean(past_stats) # past_stats is a list of stats dicts from past jobs
        std = compute_std(past_stats)
        assert mean - std <= crawled_items <= mean + std
```

rennerocha:
This was done in https://github.com/scrapinghub/spidermon/pull/137 
Now you can enable a custom StatsCollector that will store the previous stats from your spider execution, allowing you to compare it in your own monitor.

PR #137, opened 2019-04-02
Add StatsCollector that access stats from previous spider executions

rennerocha:
This should allow #122 to be done.


docs/source/stats-collection.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Comparing Spider Executions
+===========================
+
+Sometimes it is worthy to compare results from previous executions of your
+spider. For example, we should be able to track whether the number of items
+returned is lower than previous executions, which may indicate a problem.
+
+Scrapy allows you to collect stats in the form key/value, but these stats are
+lost when the spider finishes its execution. Spidermon provides a built-in
```
muzaffaryousaf: I would say those stats are set on the dash instead of lost and one have to access them explicitly via scrapinghub client. Also, stats are still accessible via spidermon `StatsMonitorMixin` but history is not available.

docs/source/stats-collection.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. code-block:: python
+
+    # myproject/monitors.py
+
+    @monitors.name("History Validation")
+    class HistoryMonitor(Monitor):
+
+      @monitors.name("Expected number of items extracted")
+      def test_expected_number_of_items_extracted(self):
```
muzaffaryousaf: can we now make this monitor part of base monitors ? We have already have a monitor to test the given number of items but it would be nice addition to check coverage based on historical crawls.

muzaffaryousaf: Separate issue could be created for this.

tests/test_extension.py
```diff
@@ -9,21 +9,6 @@
 from spidermon.contrib.scrapy.extensions import Spidermon
```
muzaffaryousaf: Let's remove unused imports `Crawler` and `Spider`

docs/source/stats-collection.rst
muzaffaryousaf: NVM, i see #122

docs/source/stats-collection.rst
Gallaecio: However, we need to take into account that Spidermon can be used outside Scrapy Cloud. I think the current text here is quite good, and the note below about Scrapy Cloud complements this text perfectly.

Maybe we could leave this text as is, but move the note about Scrapy Cloud a bit higher, maybe right before the example (before the ‘The following example…’ paragraph).

docs/source/stats-collection.rst
```diff
@@ -0,0 +1,72 @@
...
+    STATS_CLASS = (
+        "spidermon.contrib.stats.statscollectors.LocalStorageStatsHistoryCollector"
+    )
+
+    # Stores the stats of the last 10 spider execution (default=100)
+    SPIDERMON_MAX_STORED_STATS = 10
+
+The following example shows how we can verify whether the number of items
+returned in the actual spider execution is higher than 90% of the mean of items
+returned in the latest spider executions.
```
Gallaecio: Instead of ‘actual’ and ‘latest’, I would talk about ‘current’ and ‘previous’.

Although ‘actual’ (en) _can_ mean ‘actual’ (pt, es), it is usually a false friend meaning ‘real’ (pt, es). In general, ‘actual’ (pt, es) should be translated as ‘current’ (en).

Here ‘latest’ could be confused with actual/current, so ‘previous’ may be a better choice.

spidermon/contrib/stats/statscollectors.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+        if os.path.isfile(stats_location):
+            with open(stats_location, "rb") as stats_file:
+                _stats_history = pickle.load(stats_file)
+        else:
+            max_stored_stats = spider.crawler.settings.getint(
+                "SPIDERMON_MAX_STORED_STATS", default=100
+            )
+            _stats_history = deque([], maxlen=max_stored_stats)
+        spider.stats_history = _stats_history
```
Gallaecio: Regarding the preceding if-else statement, it looks to me like `SPIDERMON_MAX_STORED_STATS` is currently ignored after the first time a spider is run. Shouldn’t the `maxlen` of the `deque` loaded with `pickle` be adjusted according to the `SPIDERMON_MAX_STORED_STATS` setting as well?

If so, maybe adding a test that specifically ensures that new runs adjust the `deque.maxlen` value accordingly could prove useful.

spidermon/contrib/stats/statscollectors.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def _persist_stats(self, stats, spider):
+        spider.stats_history.appendleft(self._stats)
+        stats_location = os.path.join(
+            self._statsdir, "{}_stats_history".format(spider.name)
+        )
```
Gallaecio: I see some repetition here and in `open_spider` that may be worth avoiding. Maybe `stats_location` could be defined on the constructor, possibly making `self._statsdir` unnecessary? I think that making `statsdir` an optional parameter to the constructor as well should still allow for the tests to override the directory with their own temporary directory.

tests/contrib/stats/test_localstoragestats.py
```diff
@@ -0,0 +1,98 @@
...
+        stats_collector._statsdir = tmpdir.strpath
+        return stats_collector
+
+    return _stats_collector
+
+
+def test_spider_has_stats_history_attribute_when_opened_with_collector(
+    get_crawler, get_stats_collector
+):
+    crawler = get_crawler()
```
Gallaecio: I’m used to the pytest fixture functions defined without the `get_` prefix (e.g. `crawler` instead of `get_crawler`), and when included in the parameters of test functions they are filled with the response value already, and do not need to be called. I am not sure if I am missing something here that explains the difference.

In detail, I see here we are using:

```python
# conftest.py
@fixture
def get_crawler():
    # …

# test_localstoragestats.py
def test_something(get_crawler):
    crawler = get_crawler()
    # …
```

Where I would expect:

```python
# conftest.py
@fixture
def crawler():
    # …

# test_localstoragestats.py
def test_something(crawler):
    # …
```

tests/contrib/stats/test_localstoragestats.py
```diff
@@ -0,0 +1,98 @@
...
+    stats_collector = get_stats_collector(crawler)
+    stats_collector.open_spider(spider)
+    stats_collector.set_value("first_execution", "value")
+    stats_collector.close_spider(spider, "finished")
+
+    stats_collector = get_stats_collector(crawler)
+    stats_collector.open_spider(spider)
+
+    assert len(spider.stats_history) == 1
+    assert spider.stats_history[0] == {"first_execution": "value"}
```
Gallaecio: :lipstick: You could use a single line here:

```python
assert spider.stats_history == [{"first_execution": "value"}]
```

Same in some tests below.

spidermon/contrib/stats/statscollectors.py
raphapassini: I also think we can move that bit of code to a separated function like `get_stats_location`, this way is also easy to overload this in the future if needed.

docs/source/stats-collection.rst
rennerocha: Much better. Thanks!

tests/test_extension.py
rennerocha: Fixed!

docs/source/stats-collection.rst
rennerocha: We can't consider that the user is running their spiders in a specific platform, so we should not include mentions to Scrapy Could (or Dash in your comment) in the main documentation (for Scrapy Cloud users, the Note block should be enough).

I moved the note block to before the sample monitor, but it didn't look good to me. I'd prefer to keep as it is now.

spidermon/contrib/stats/statscollectors.py
rennerocha: It makes sense. I updated the code and the tests for this scenario.

tests/contrib/stats/test_localstoragestats.py
rennerocha: I wanted the fixture to be a factory function for a Crawler instance because for some tests I want to send some extra arguments for the Crawler initialization.

spidermon/contrib/stats/statscollectors.py
rennerocha: I removed the code duplication. It makes no sense to add it to the constructor as it will never be called except for the tests.

tests/contrib/stats/test_localstoragestats.py
Gallaecio: A factory function makes a lot of sense. But then I am not sure what’s the benefit of using a fixture. Wouldn’t it work the same if instead of a fixture returning a function you simply define that function and call it from tests?

docs/source/stats-collection.rst
```diff
@@ -0,0 +1,71 @@
...
+          self.assertFalse(
+              items_extracted <= minimum_threshold,
+              msg="Expected at least {} items extracted.".format(
+                  minimum_threshold
+              ),
+          )
+
+.. note::
+    If you are running your spider in `Scrapy Cloud`_ you need to enable the
+    `DotScrapy Persistence Add-on`_ in your project to keep your `.scrapy` directory
```
Gallaecio: :lipstick: 

```suggestion
    `DotScrapy Persistence Add-on`_ in your project to keep your ``.scrapy`` directory
```

spidermon/contrib/stats/statscollectors.py
```diff
@@ -0,0 +1,39 @@
...
+
+        if os.path.isfile(self.stats_location):
+            with open(self.stats_location, "rb") as stats_file:
+                _stats_history = pickle.load(stats_file)
+        else:
+            _stats_history = deque([], maxlen=max_stored_stats)
+
+        if _stats_history.maxlen != max_stored_stats:
+            _stats_history = deque([
+                stat for stat in _stats_history
```
Gallaecio: I believe you can simply use `deque(_stats_history, maxlen=max_stored_stats)`.

Gallaecio: I only tested in Python 3.6, though, maybe older versions do not support that.

rennerocha: You are right. It worked in Python 2.7 as well.

PR #138, opened 2019-04-05
docs: Provide API documentation for render_text_template and render_template

### LINKED ISSUE #124: What is the difference between render_text_template and render_template?
Hi,

I would like to know what is the intention of `render_text_template` and `render_template` methods in `/contrib/actions/templates.py`. 

https://github.com/scrapinghub/spidermon/blob/47b052b86b2ad1197f87eb37f4ce21a7c2e7516b/spidermon/contrib/actions/templates.py#L37-L43

both are the same.

Gallaecio:
Fixes #124


vipulgupta2048:
I think this looks good to go.
CC:@matiskay does this resolve your issue? In reference to #124

matiskay:
@vipulgupta2048 yes, it does.

raphapassini:
In general I think that the template deserves a topic in our official documentation.
Maybe we can improve this in the next release, but now at least we have the code documented.

Very nice work :)

Issue #51, opened 2018-10-25
Adding new JSON schema format validators


andrewbaxter:
We're exporting to SQL and having `datetime` objects in the data is the easiest way to send SQL datetimes, but these are rejected by jsonschema.  There's no way to set a validator class or pass extra arguments to the current validator class to override the datetime validator AFAICT.

manycoding:
@andrewbaxter Could you elaborate?
`"format": "date-time"` is probably what you need.

The JSON schema standard supports `formats` and the library used here requires additional dependencies to make them work - jsonschema[formats], see here https://github.com/Julian/jsonschema/blob/master/docs/validate.rst#validating-formats

And there's an official standard https://json-schema.org/latest/json-schema-validation.html#rfc.section.7.3

Just to say, there's `fastjsonschema` which has format validation by default without additional tweaks.

Gallaecio:
@andrewbaxter Should we close this then?

andrewbaxter:
Sorry for the delay - @manycoding and thanks for pinging me @Gallaecio .

I tried "format": "date-time" and it doesn't seem to do anything, ex:
Schema:
```
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
   "properties": {
     "v": {
       "type": "string",
       "format": "date-time"
     }
   }
 }
```

Shows:
```
2019-04-05 20:33:27 [scrapy.core.scraper] DEBUG: Scraped from <200 https://scrapinghub.com/>
{'v': 4, '_validation': defaultdict(<class 'list'>, {'v': ['Invalid string']})}
2019-04-05 20:33:27 [scrapy.core.scraper] DEBUG: Scraped from <200 https://scrapinghub.com/>
{'v': 'hi'}
```
The first just to show that validation is happening.  TBH I'm not sure I did that correctly, but if this worked I think that would be very close to what I want.

To clarify on this though, even if Spidermon supports the builtin datetime there are still a lot of types that that are not supported:
```
        yield {
            'v': arrow.now(),
        }
        yield {
            'f': decimal.Decimal('1.03'),
        }
        yield {
            'u': uuid.uuid4(),
        }
```

results in
```
2019-04-05 20:38:20 [scrapy.core.scraper] ERROR: Error processing {'v': <Arrow [2019-04-05T20:38:20.617099+09:00]>}
Traceback (most recent call last):
[...]

PR #143, opened 2019-04-08
[+1 MRG] Update setup.py

manycoding:
I propose to include `format` since it seems an obvious mistake that people use `format` and miss the dependencies.
The problem is if you have `format` validator and lack deps you won't get any error and probably will never know.
Related to #51


raphapassini:
LGTM

PR #147, opened 2019-04-10
Update setup.py

manycoding:
@raphapassini I forgot to remove strict-rfc3339 dep. It's in jsonschema[format] and is only used with jsonschema validation.
#143


manycoding:
@vipulgupta2048 Good point, I closed #114 as obsolete.

PR #150, opened 2019-04-15
Adding a monitor to match a expected item count

ejulio:
It is common for our spiders to have some kind of seed URLs to crawl the data from.
This monitor enables us to match this kind of seed with `item_scraped_count`.
We only need to set a stats `spidermon/expected_item_scraped_count` and it'll do the magic.
Also, we can make it a bit flexible by setting `SPIDERMON_EXPECTED_ITEMS_THRESHOLD` to let it miss some items (maybe bans, invalid pages, redirects, ...).

What are your thoughts about it?

### DIFF
--- a/spidermon/contrib/scrapy/monitors.py
+++ b/spidermon/contrib/scrapy/monitors.py
@@ -44,6 +44,38 @@ class ItemCountMonitor(BaseScrapyMonitor):
         self.assertTrue(item_extracted >= self.minimum_threshold, msg=msg)
 
 
+@monitors.name("Extracted Expected Items Monitor")
+class ExpectedItemCountMonitor(BaseScrapyMonitor):
+    """Check if spider extracted the expected minimum number of items provided
+    in a stats `spidermon/expected_item_scraped_count`.
+
+    You can configure it using ``SPIDERMON_EXPECTED_ITEMS_THRESHOLD`` setting.
+    The default value is 0, meaning that the number of scraped items should be
+    equal to the expected one.
+    Any other value in the range [0, 1] will be computed as a percentage
+    threshold and subtracted from `spidermon/expected_item_scraped_count`.
+    """
+
+    def run(self, result):
+        settings = self.crawler.settings
+        self.expected_items_threshold = settings.getfloat('SPIDERMON_EXPECTED_ITEMS_THRESHOLD', 0)
+        if not (0.0 <= self.expected_items_threshold <= 1.0):
+            raise NotConfigured("You must specify a threshold of expected items.")
+        return super(ExpectedItemCountMonitor, self).run(result)
+
+    @monitors.name("Should extract the expected amount of items")
+    def test_expected_amount_of_items(self):
+        item_scraped_count = self.stats.get("item_scraped_count", 0)
+        expected_item_scraped_count = self.stats.get("spidermon/expected_item_scraped_count", 0)
+        threshold = int(expected_item_scraped_count * self.expected_items_threshold)
+        expected_item_scraped_count -= threshold
+
+        msg = "Extracted {} items, expected at least {}".format(
+            item_scraped_count, expected_item_scraped_count
+        )
+        self.assertTrue(item_scraped_count >= expected_item_scraped_count, msg=msg)
+
+
 @monitors.name("Error Count Monitor")
 class ErrorCountMonitor(BaseScrapyMonitor):
     """Check for errors in the spider log.
--- a/tests/test_monitors.py
+++ b/tests/test_monitors.py
@@ -5,6 +5,7 @@ from scrapy.crawler import Crawler
 from spidermon.contrib.scrapy.runners import SpiderMonitorRunner
 from spidermon.contrib.scrapy.monitors import (
     FinishReasonMonitor,
+    ExpectedItemCountMonitor,
     ItemCountMonitor,
     ErrorCountMonitor,
     UnwantedHTTPCodesMonitor,
@@ -36,6 +37,11 @@ def make_data(request):
 @pytest.fixture
 def item_count_suite():
     return MonitorSuite(monitors=[ItemCountMonitor])
+
+
+@pytest.fixture
+def expected_item_count_suite():
+    return MonitorSuite(monitors=[ExpectedItemCountMonitor])
 
 
 def new_suite(monitors):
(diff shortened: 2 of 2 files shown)


spidermon/contrib/scrapy/monitors.py
```diff
@@ -44,6 +44,38 @@ def test_minimum_number_of_items(self):
         self.assertTrue(item_extracted >= self.minimum_threshold, msg=msg)
 
 
+@monitors.name("Extracted Expected Items Monitor")
+class ExpectedItemCountMonitor(BaseScrapyMonitor):
```
ejulio: Lookig for a better name.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -44,6 +44,38 @@ def test_minimum_number_of_items(self):
...
+    """Check if spider extracted the expected minimum number of items provided
+    in a stats `spidermon/expected_item_scraped_count`.
+
+    You can configure it using ``SPIDERMON_EXPECTED_ITEMS_THRESHOLD`` setting.
```
ejulio: Looking for a better name for this setting also.
Moreover, not sure it should a float [0, 1] or an in [0, 100].

One example to think here Say that I allow 10% of missed items.
Is it better:
* `SPIDERMON_EXPECTED_ITEMS_THRESHOLD = 0.1`
* `SPIDERMON_EXPECTED_ITEMS_THRESHOLD = 0.9`

tests/test_monitors.py
```diff
@@ -91,6 +97,69 @@ def test_item_count_monitor_extracted_less_than_expected(make_data, item_count_s
...
+    data = make_data({"SPIDERMON_EXPECTED_ITEMS_THRESHOLD": 0.0})
+    runner = data.pop("runner")
+    data["stats"]["item_scraped_count"] = 500
+    data["stats"]["spidermon/expected_item_scraped_count"] = 500
+    runner.run(expected_item_count_suite, **data)
+    assert runner.result.monitor_results[0].error is None
+
+
+def test_expected_item_count_monitor_no_expected_count(make_data, expected_item_count_suite):
+    """ExpectedItemCountMonitor should pass when extract the expected number of items """
```
ejulio: I wanted to skip the monitor, not make it pass (I think the semantics are different).
However, currently, it is not possible to skip a monitor.
Probably, I'll work in another PR for that and then change it here

Gallaecio:
I wonder if, instead of creating a separate monitor, it would make sense to extend the existing `ItemCountMonitor` monitor with this functionality.

ejulio:
I have the same question @Gallaecio :stuck_out_tongue: 
My thought to create a new one was that `ItemCountMonitor` raises as `NotConfigured` if the minimum number of items is not set. Then, there would be conditions to raise `NotConfigured` for the `ExpectedItemCountMonitor` and `ItemCountMonitor`.
Basically, it wouldn't be so obvious which monitor was enabled.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -44,6 +44,38 @@ def test_minimum_number_of_items(self):
...
+    equal to the expected one.
+    Any other value in the range [0, 1] will be computed as a percentage
+    threshold and subtracted from `spidermon/expected_item_scraped_count`.
+    """
+
+    def run(self, result):
+        settings = self.crawler.settings
+        self.expected_items_threshold = settings.getfloat('SPIDERMON_EXPECTED_ITEMS_THRESHOLD', 0)
+        if not (0.0 <= self.expected_items_threshold <= 1.0):
+            raise NotConfigured("You must specify a threshold of expected items.")
```
manycoding: For me both ItemCount monitors look the same, so considering https://github.com/scrapinghub/spidermon/pull/150#issuecomment-484024765, what if we:
1. Remove `SPIDERMON_MIN_ITEMS` and only rely on `expected_item_scraped_count`
2. Set the default threshold to expect all expected items, meaning we don't raise not configured
So it's only the matter of writing `expected_item_scraped_count`

Of even better, let's forget about `expected_item_scraped_count` and use only thresholds to check if we scraped items based on`item_scraped_count` * threshold

spidermon/contrib/scrapy/monitors.py
manycoding: I think more obvious would be `1` if we expect all `expected_item_scraped_count` and `0` if we expect 0.
So `SPIDERMON_EXPECTED_ITEMS_THRESHOLD = 0.9`

spidermon/contrib/scrapy/monitors.py
```diff
@@ -44,6 +44,38 @@ def test_minimum_number_of_items(self):
...
+    The default value is 0, meaning that the number of scraped items should be
+    equal to the expected one.
+    Any other value in the range [0, 1] will be computed as a percentage
+    threshold and subtracted from `spidermon/expected_item_scraped_count`.
+    """
+
+    def run(self, result):
+        settings = self.crawler.settings
+        self.expected_items_threshold = settings.getfloat('SPIDERMON_EXPECTED_ITEMS_THRESHOLD', 0)
+        if not (0.0 <= self.expected_items_threshold <= 1.0):
```
manycoding: Why we could expect 0 items?

spidermon/contrib/scrapy/monitors.py
```diff
@@ -44,6 +44,38 @@ def test_minimum_number_of_items(self):
...
+    You can configure it using ``SPIDERMON_EXPECTED_ITEMS_THRESHOLD`` setting.
+    The default value is 0, meaning that the number of scraped items should be
+    equal to the expected one.
+    Any other value in the range [0, 1] will be computed as a percentage
+    threshold and subtracted from `spidermon/expected_item_scraped_count`.
+    """
+
+    def run(self, result):
+        settings = self.crawler.settings
+        self.expected_items_threshold = settings.getfloat('SPIDERMON_EXPECTED_ITEMS_THRESHOLD', 0)
```
manycoding: So if we set it to 0, how can NotConfigured be ever raised?

tests/test_monitors.py
```diff
@@ -91,6 +97,69 @@ def test_item_count_monitor_extracted_less_than_expected(make_data, item_count_s
     )
 
 
+def test_needs_to_configure_expected_item_count_monitor(make_data, expected_item_count_suite):
+    """Should raise an exception when ExpectedItemCountMonitor it's not configured """
+    data = make_data({'SPIDERMON_EXPECTED_ITEMS_THRESHOLD': 1.1})
```
manycoding: So it's not `NotConfigured`, it's `ValueError`

tests/test_monitors.py
```diff
@@ -91,6 +97,69 @@ def test_item_count_monitor_extracted_less_than_expected(make_data, item_count_s
...
+
+
+def test_expected_item_count_monitor_less_than_expected(make_data, expected_item_count_suite):
+    """ExpectedItemCountMonitor should fail when the desired # of items is not extracted """
+    data = make_data({'SPIDERMON_EXPECTED_ITEMS_THRESHOLD': 0.1})
+    runner = data.pop("runner")
+    data["stats"]["item_scraped_count"] = 10
+    data["stats"]["spidermon/expected_item_scraped_count"] = 20
+    runner.run(expected_item_count_suite, **data)
+    for r in runner.result.monitor_results:
```
manycoding: It looks like it's a list of size 1, so why we just don't get the first one instead of using `for`

tests/test_monitors.py
```diff
@@ -91,6 +97,69 @@ def test_item_count_monitor_extracted_less_than_expected(make_data, item_count_s
...
+        assert "Extracted 10 items, expected at least 18" in r.error
+
+
+def test_expected_item_count_monitor_more_than_expected(make_data, expected_item_count_suite):
```
manycoding: Notice how you repeat the code, I would use `@pytest.parametrize` like here https://github.com/scrapinghub/arche/blob/60c8aa87a67b432ec1fa0c9690d43684373b5d53/tests/tools/test_api.py#L7 instead of having these 4 separate repeating tests.

rennerocha:
`spidermon.contrib.scrapy.monitor.ItemCountMonitor` is almost exactly the same as your new monitor, so I am quite unsure if we should add a new basic monitor just for your specific scenario.

Also the main objective to include built-in monitors in spidermon was to make easier for an user to include basic monitoring in their spiders using only a set of new settings in the project. You monitor requires a non-default stat (spidermon/expected_item_scraped_count) to be provided, so it is not a plug-n-play monitor.

I believe that this monitor should not be included in the set of built-in monitors.

ejulio:
@rennerocha , I agree on the matter of requiring a non-standard stats for that.
Do you you there is a better way to add it?

Actually, if scrapy settings were updatable after the spider started, I think we could set the `ItemCountMonitor` minimum count dynamically.

spidermon/contrib/scrapy/monitors.py
ejulio: It is not that we expect 0, we expected scraped_count - scraped_count * threshold.
If threshold is 0, then we expect scraped_count.

spidermon/contrib/scrapy/monitors.py
ejulio: That makes.
Just the last case, I think I didn't get it. Mostly because against what am I going to compare `scraped_count * threshold`?

ejulio: Also, removing the settings breaks backward compatibility...

manycoding:
@rennerocha @ejulio That's a good point, so why not use the default state as I pointed at the end here https://github.com/scrapinghub/spidermon/pull/150/files#r276289610 ?

spidermon/contrib/scrapy/monitors.py
manycoding: Why do we need backwards compatibility?
I confused myself since we need the amount of scraped items and the amount of expected items as well.

ejulio: Because if there is someone using the library we cannot starting throwing errors when they update.
At first, we need to notify them that the API will change in the following release or so

manycoding:
>spidermon.contrib.scrapy.monitor.ItemCountMonitor is almost exactly the same as your new monitor, so I am quite unsure if we should add a new basic monitor just for your specific scenario.

Spiders which scrape from provided input file is rather common, I wouldn't call it specific at all. But I agree that it could be just one monitor for both cases (when we want to verify that the scraped items more or less the same as the number of provided items, and when we want to verify just the number of items, provided it stays the same)

>Also the main objective to include built-in monitors in spidermon was to make easier for an user to include basic monitoring in their spiders using only a set of new settings in the project. You monitor requires a non-default stat (spidermon/expected_item_scraped_count) to be provided, so it is not a plug-n-play monitor.
I believe that this monitor should not be included in the set of built-in monitors.

So let's create a new category then which fits the needs so as to not implement this solution again and again?

spidermon/contrib/scrapy/monitors.py
manycoding: So let's notify them then :) Anyway, let's figure out if we want to keep it separate, make it a basic monitor first.

rennerocha: We need to keep things  backward compatible as much as possible to avoid projects break after version updates. We have no idea who is using the library, so it is impossible to notify all of them. A good library will not break unless it is totally necessary, and this is not a change that would improve the library to justify it.

rennerocha:
> @rennerocha , I agree on the matter of requiring a non-standard stats for that.
> Do you you there is a better way to add it?
> 
> Actually, if scrapy settings were updatable after the spider started, I think we could set the `ItemCountMonitor` minimum count dynamically.

Unfortunately it is not possible to update Scrapy settings after the spider started. There is a  [update_settings](https://github.com/scrapy/scrapy/issues/2392#issuecomment-259661978) method but it doesn't have access to spider input arguments.

This is a limitation of Scrapy as mentioned [here](https://github.com/scrapy/scrapy/issues/3663) and [here](https://github.com/scrapy/scrapy/issues/2392). Probably when [this issue](https://github.com/scrapy/scrapy/issues/1305) is solved we can do something about it.

rennerocha:
> @rennerocha
> 
> > spidermon.contrib.scrapy.monitor.ItemCountMonitor is almost exactly the same as your new monitor, so I am quite unsure if we should add a new basic monitor just for your specific scenario.
> 
> Spiders which scrape from provided input file is rather common, I wouldn't call it specific at all. But I agree that it could be just one monitor for both cases (when we want to verify that the scraped items more or less the same as the number of provided items, and when we want to verify just the number of items, provided it stays the same)

I can see now that this is a common pattern and it would be nice to have an easy solution for this (probably with one monitor for both cases). However I believe that this is out-of-scope of Spidermon built-in monitors (as it requires non-default stats to be set, instead of just including a new setting).

> > Also the main objective to include built-in monitors in spidermon was to make easier for an user to include basic monitoring in their spiders using only a set of new settings in the project. You monitor requires a non-default stat (spidermon/expected_item_scraped_count) to be provided, so it is not a plug-n-play monitor.
> > I believe that this monitor should not be included in the set of built-in monitors.
> 
> So let's create a new category then which fits the needs so as to not implement this solution again and again?

I can't see how to include a new category. Maybe include an example of this scenario in the docs in a Cookbook section?

spidermon/contrib/scrapy/monitors.py
manycoding: Maybe it's not the change to break backwards compatibility.
All good libraries I am familiar with have deprecations so they still can change fast and have flexibility. Numpy, pandas, fastai to name a few.

ejulio:
@rennerocha , @victor-torres and @manycoding .
I'll close this PR since it makes more sense to update the settings as pointed out by @rennerocha.
Thanks for your thoughts and comments

PR #152, opened 2019-04-22
Use setup reqs in tox, remove redundant mock

manycoding:


manycoding:
This pr
1. Makes sure tox uses requirements as specified in setup.py (not from other files)
2. Fixes links to these reqs in tox.ini
3. Cleans useless files

PR #155, opened 2019-05-03
email_subject*: don't fail if SPIDERMON_EMAIL_SUBJECT_TEMPLATE is configured in favor of SPIDERMON_EMAIL_SUBJECT

Toritrae:
Allow usage of SPIDERMON_EMAIL_SUBJECT_TEMPLATE instead of SPIDERMON_EMAIL_SUBJECT without failing during initialization. If EMAIL_SUBJECT is configured it still has precedence.

### DIFF
--- a/spidermon/contrib/actions/email/__init__.py
+++ b/spidermon/contrib/actions/email/__init__.py
@@ -59,7 +59,7 @@ class SendEmail(ActionWithTemplates):
             raise NotConfigured(
                 "You must provide at least one recipient for the message."
             )
-        if not self.subject:
+        if not self.subject and not self.subject_template:
             raise NotConfigured("You must provide a subject for the message.")
         if not (
             self.body_text or body_text_template or body_html or self.body_html_template


spidermon/contrib/actions/email/__init__.py
```diff
@@ -59,7 +59,7 @@ def __init__(
             raise NotConfigured(
                 "You must provide at least one recipient for the message."
             )
-        if not self.subject:
+        if not self.subject and not self.subject_template:
             raise NotConfigured("You must provide a subject for the message.")
```
Gallaecio: What about updating the message in `NotConfigured` to say `a subject or a subject template`?

Toritrae:
```suggestion
            raise NotConfigured("You must provide a subject or a subject-template for the message.")
```
Reason I did not include that was a similar message for the setting the body: "you must provide a body for...". But your suggestion makes sense of course.
Wouldn't it make sense to use the names of the settings here? Else it's necessary to guess/look-up the name of the setting. But I'm fine with either one tbh.

It's my first time using a suggestion, is this enough to pick up the change and resolve this?

Gallaecio: I also like the idea of including the setting names, for example between parentheses after each concept (subject, subject template).

By the way, I don’t think ‘subject-template’ is necessary or correct English. [Hyphens can be used for joining words](https://en.wikipedia.org/wiki/Hyphen#Joining), but I don’t think this is the right context for that. It would be correct however for ‘subject-defining template’ or ‘subject-*templater*’.

As for suggestions, you can pick the change to quickly apply it to your branch, but you can also simply perform the change yourself on a new commit.

Gallaecio:
:+1: by me.

I must say I’m surprised it passes all checks, though :slightly_smiling_face: . I think Spidermon uses Black, and Black line length limit is 88.

Toritrae:
I played around with line length a bit, seems it makes no difference to black. I reformatted the message regardless.

spidermon/contrib/actions/email/__init__.py
rennerocha: I liked the idea of including the related spidermon setting inside parenthesis in this message, so the error message will be much more helpful. I believe that `subject template` (without hyphen) is the right way to mention this setting.

rennerocha:
I know that we don't have tests for the actions (and it is something we need to improve), so would you mind create in this PR a new test file for this action and add some tests to make sure that the lack of one of the subject arguments raise an exception? Other tests for other arguments may be created after that (I don't want a PR handles more than one feature).

Toritrae:
No problem, will be some time before I get a shot at this though.

rennerocha:
@Toritrae , would you be able to include the tests for your code?

Tariq187:
Visa

Issue #153, opened 2019-04-23
Create a default setting for SPIDERMON_UNWANTED_HTTP_CODES


ejulio:
Say that I want to increase the default threshold for `SPIDERMON_UNWANTED_HTTP_CODES`.

Currently, I need to copy/paste
```
'SPIDERMON_UNWANTED_HTTP_CODES': {
    code: 100 for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
}
```

1. Maybe we can create a base setting, like https://github.com/scrapy/scrapy/blob/master/scrapy/settings/default_settings.py#L93.
This way, I can iterate over this dict instead of copy/pasting the error codes.

2. Maybe we can create a new setting for the default threshold. This would this setting instead of the const 10 for the default error codes.

Other ideas?

My preference is for 2. It looks cleaner, easier to setup and implement.

rosheen33:
Under Progress:
https://github.com/scrapinghub/spidermon/compare/default_threshold?expand=1

rosheen33:
PR For Review:
https://github.com/scrapinghub/spidermon/pull/156

PR #156, opened 2019-05-03
Added New Setting "SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD" [#153]

rosheen33:
Issue:
https://github.com/scrapinghub/spidermon/issues/153

### DIFF
--- a/spidermon/contrib/scrapy/monitors.py
+++ b/spidermon/contrib/scrapy/monitors.py
@@ -87,17 +87,19 @@ class FinishReasonMonitor(BaseScrapyMonitor):
 @monitors.name("Unwanted HTTP codes monitor")
 class UnwantedHTTPCodesMonitor(BaseScrapyMonitor):
     """Check for maximum number of unwanted HTTP codes.
+    You can configure it using ``SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT`` setting.
+    or ``SPIDERMON_UNWANTED_HTTP_CODES`` setting
 
     Usage # 1
 
-    You can configure an ``int`` of maximum unwanted HTTP codes with
+    You can configure a setting ``int`` of maximum unwanted HTTP codes with
     ``SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT`` the default value is::
 
         SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT = 10
 
     Usage # 2
 
-    You can configure a ``dict`` or ``list`` of unwanted HTTP codes with
+    You can configure a setting ``dict`` or ``list`` of unwanted HTTP codes with
     ``SPIDERMON_UNWANTED_HTTP_CODES`` the default value is::
 
         1.1 - In case of Dictionary:
--- a/spidermon/contrib/scrapy/monitors.py
+++ b/spidermon/contrib/scrapy/monitors.py
@@ -6,6 +6,7 @@ SPIDERMON_MIN_ITEMS = "SPIDERMON_MIN_ITEMS"
 SPIDERMON_MAX_ERRORS = "SPIDERMON_MAX_ERRORS"
 SPIDERMON_EXPECTED_FINISH_REASONS = "SPIDERMON_EXPECTED_FINISH_REASONS"
 SPIDERMON_UNWANTED_HTTP_CODES = "SPIDERMON_UNWANTED_HTTP_CODES"
+SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD = "SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD"
 
 
 class BaseScrapyMonitor(Monitor, SpiderMonitorMixin):
@@ -87,24 +88,28 @@ class FinishReasonMonitor(BaseScrapyMonitor):
 class UnwantedHTTPCodesMonitor(BaseScrapyMonitor):
     """Check for maximum number of unwanted HTTP codes.
 
-    You can configure a ``dict`` of unwanted HTTP codes with
-    ``SPIDERMON_UNWANTED_HTTP_CODES`` the default value is::
+    You can configure a ``list`` of unwanted HTTP codes with
+    ``SPIDERMON_UNWANTED_HTTP_CODES`` and ``SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD``
+     the default values are::
 
-        DEFAULT_ERROR_CODES = {
-            code: 10
-            for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]}
+        DEFAULT_ERROR_CODES = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
+        DEFAULT_ERROR_THRESHOLD = 10
     """
 
-    DEFAULT_ERROR_CODES = {
-        code: 10 for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
-    }
+    DEFAULT_ERROR_CODES = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
+    DEFAULT_ERROR_THRESHOLD = 10
 
     @monitors.name("Should not hit the limit of unwanted http status")
     def test_check_unwanted_http_codes(self):
         error_codes = self.crawler.settings.getdict(
             SPIDERMON_UNWANTED_HTTP_CODES, self.DEFAULT_ERROR_CODES
         )
-        for code, max_errors in error_codes.items():
+        errors_threshold = self.crawler.settings.getdict(
+            SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD, self.DEFAULT_ERROR_THRESHOLD
+        )
+        error_codes_dict = {code: errors_threshold for code in error_codes}
+
+        for code, max_errors in error_codes_dict.items():
             code = int(code)
             count = self.stats.get(
                 "downloader/response_status_count/{}".format(code), 0
--- a/spidermon/contrib/scrapy/monitors.py
+++ b/spidermon/contrib/scrapy/monitors.py
@@ -95,13 +95,6 @@ class UnwantedHTTPCodesMonitor(BaseScrapyMonitor):
 
         SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT = 10
 
-        ----->
-
-        SPIDERMON_UNWANTED_HTTP_CODES = {
-            code: SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT
-            for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
-        }
-
     Usage # 2
 
     You can configure a ``dict`` or ``list`` of unwanted HTTP codes with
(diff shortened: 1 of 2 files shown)


Gallaecio:
I think this is not what #153 suggested. The approach suggested in #153 is much simpler: just allow defining a threshold for all default codes in `SPIDERMON_UNWANTED_HTTP_CODES`, so that changing the threshold without adding or removing codes from `SPIDERMON_UNWANTED_HTTP_CODES` does not require overriding the default `SPIDERMON_UNWANTED_HTTP_CODES`.

But `SPIDERMON_UNWANTED_HTTP_CODES` should stay the same otherwise: continue being a dict. If `SPIDERMON_UNWANTED_HTTP_CODES` is overridden by the user, `SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD` would simply have no effect.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -6,6 +6,7 @@
 SPIDERMON_MAX_ERRORS = "SPIDERMON_MAX_ERRORS"
 SPIDERMON_EXPECTED_FINISH_REASONS = "SPIDERMON_EXPECTED_FINISH_REASONS"
 SPIDERMON_UNWANTED_HTTP_CODES = "SPIDERMON_UNWANTED_HTTP_CODES"
+SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD = "SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD"
```
ejulio: Don't knoe if threshold is good name, maybe `SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT`.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,24 +88,28 @@ def test_should_finish_with_expected_reason(self):
...
-    DEFAULT_ERROR_CODES = {
-        code: 10 for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
-    }
+    DEFAULT_ERROR_CODES = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
+    DEFAULT_ERROR_THRESHOLD = 10
 
     @monitors.name("Should not hit the limit of unwanted http status")
     def test_check_unwanted_http_codes(self):
-        error_codes = self.crawler.settings.getdict(
+        error_codes = self.crawler.settings.getlist(
```
ejulio: This breaks backwards compatibility.
Maybe:
* try to read a dict (keeps it backwards compatible and enables us to configure different counts for each error)
* If it is a dict, just keep it like that
* If it is a list, create a dict using the codes in the list and `SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLD`

Gallaecio: Oh, supporting both lists and dicts, great idea.

As part of https://github.com/scrapy/scrapy/pull/3696 Scrapy’s `Setting` class is getting a `getdictorlist` method, but since you probably do not want to wait for that to be merged or have Spidermon require some future Scrapy version, you might need to hardcode something in the lines of `getdictorlist` here.

rosheen33:
@ejulio We also need to update the Documentation after adding this new feature ?

rosheen33:
@ejulio @Gallaecio Please have a look again
I have added backward compatibility.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,24 +88,48 @@ def test_should_finish_with_expected_reason(self):
...
 
-    DEFAULT_ERROR_CODES = {
-        code: 10 for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
-    }
+    DEFAULT_ERROR_CODES = {code: 10 for code in ERROR_CODES}
 
     @monitors.name("Should not hit the limit of unwanted http status")
     def test_check_unwanted_http_codes(self):
-        error_codes = self.crawler.settings.getdict(
+        error_codes_dict = self.crawler.settings.getdict(
```
ejulio: Can't we read a list as well?
My idea is that, I can configure an arbitrary dict.
Or I can use the default _error count with some custom HTTP codes_, or use a _custom error count with default HTTP codes_.

tests/test_monitors.py
```diff
@@ -154,7 +155,20 @@ def test_unwanted_httpcodes_should_fail(make_data):
 def test_unwanted_httpcodes_should_pass(make_data):
     """Unwanted HTTP Code should pass if # off responses with error status
     codes is lower than expected """
+
+    # Scenario # 1
+    data = make_data({SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT: 16})
```
ejulio: What do you think of covering the same scenarios in `test_unwanted_httpcodes_should_fail`?

ejulio:
> @ejulio We also need to update the Documentation after adding this new feature ?

@rosheen33, as far as I remember, the docs are built from docstrings, so your comments in the PR should do the job.

spidermon/contrib/scrapy/monitors.py
rosheen33: Yes I thought about that too.
But I think that would be kind of messy.
If the user wants to change the default error codes.
He can do that through dictionary setting.

rosheen33:
> > @ejulio We also need to update the Documentation after adding this new feature ?
> 
> @rosheen33, as far as I remember, the docs are built from docstrings, so your comments in the PR should do the job.
That will be great 👍

rosheen33:
@ejulio I have added dict and list compatibility
Also added all the scenarios covered in test cases.
Please have a look.

rosheen33:
@ejulio How many reviews are required ?
Am I allowed to merge after one approval ?

ejulio:
I think we can wait @rennerocha for merging.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,23 +88,60 @@ def test_should_finish_with_expected_reason(self):
...
+        }
+
+    """
+    DEFAULT_ERROR_MAX_COUNT = 10
+    DEFAULT_ERROR_CODES = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
 
     @monitors.name("Should not hit the limit of unwanted http status")
     def test_check_unwanted_http_codes(self):
-        error_codes = self.crawler.settings.getdict(
+        error_codes = self.crawler.settings.get(
```
Gallaecio: The problem with using `get` here is that it will not be possible to set this option from the command line, as the value will always be a string.

Please, look at my implementation of `getdictorlist` at scrapy/scrapy#3696, and consider using a similar implementation here to read the setting as either a dictionary or a list.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,23 +88,60 @@ def test_should_finish_with_expected_reason(self):
...
+
+        ----->
+
+        SPIDERMON_UNWANTED_HTTP_CODES = {
+            code: SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT / 10
+            for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
+        }
+
+    """
+    DEFAULT_ERROR_MAX_COUNT = 10
```
rennerocha: Maybe these are best names for this attributes:
```
DEFAULT_UNWANTED_HTTP_CODES = 10
DEFAULT_UNWANTED_HTTP_CODES_MAX_COUNT = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
```
They keep almost the same name as the setting to modify the default values, which could make it easier for maintenance later.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,23 +88,60 @@ def test_should_finish_with_expected_reason(self):
 class UnwantedHTTPCodesMonitor(BaseScrapyMonitor):
     """Check for maximum number of unwanted HTTP codes.
 
-    You can configure a ``dict`` of unwanted HTTP codes with
+    Usage # 1
```
rennerocha: The documentation for this monitor is generated automatically from the docstrings, but I'd suggest that you build the docs and check if they look good. Install the [docs requirements](https://github.com/scrapinghub/spidermon/blob/master/requirements-docs.txt) in your project, and [build the docs from this directory](https://github.com/scrapinghub/spidermon/tree/master/docs) using `make html`.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,23 +88,60 @@ def test_should_finish_with_expected_reason(self):
...
+
+    You can configure an ``int`` of maximum unwanted HTTP codes with
+    ``SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT`` the default value is::
+
+        SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT = 10
```
rennerocha: It is not clear where the user needs to change this values, neither what the section after the `----->` means.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,23 +88,60 @@ def test_should_finish_with_expected_reason(self):
...
+            for code in [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
+        }
+
+    """
+    DEFAULT_ERROR_MAX_COUNT = 10
+    DEFAULT_ERROR_CODES = [400, 407, 429, 500, 502, 503, 504, 523, 540, 541]
 
     @monitors.name("Should not hit the limit of unwanted http status")
     def test_check_unwanted_http_codes(self):
-        error_codes = self.crawler.settings.getdict(
```
rennerocha: Instead of `error_codes` I would use `unwanted_http_codes` which shows more clearly what we are trying to verify in this monitor (not all codes are errors). Maybe change it in the whole monitor would be good.

spidermon/contrib/scrapy/monitors.py
rennerocha: I agree with @Gallaecio . I don't see any problem using part of his code instead of using a simple `get`.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -87,23 +88,60 @@ def test_should_finish_with_expected_reason(self):
...
+
+        errors_max_count = self.crawler.settings.getint(
+            SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT, self.DEFAULT_ERROR_MAX_COUNT
+        )
+
+        if not isinstance(error_codes, dict):
+            error_codes = {code: errors_max_count for code in error_codes}
+
         for code, max_errors in error_codes.items():
             code = int(code)
```
rennerocha: I believe that the codes are always integers (make this clear in the docs, BTW), so we don't need to convert it here.

spidermon/contrib/scrapy/monitors.py
rosheen33:
```
  File "Documents/pyenv/spidermon/bin/sphinx-build", line 10, in <module>
    sys.exit(main())
  File "Documents/pyenv/spidermon/lib/python3.6/site-packages/sphinx/cmd/build.py", line 313, in main
    locale.setlocale(locale.LC_ALL, '')
  File "Documents/pyenv/spidermon/lib/python3.6/locale.py", line 598, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting
```
Getting error on the following command.

spidermon/contrib/scrapy/monitors.py
rosheen33: part after `---->`
Shows where the value mentioned will be used.

spidermon/contrib/scrapy/monitors.py
rosheen33: @Gallaecio Do I need to wait for this branch to merge ?
then I can use it Right ?

spidermon/contrib/scrapy/monitors.py
rosheen33: @rennerocha I have not touched this part of code. It was there pre hand.

spidermon/contrib/scrapy/monitors.py
rosheen33: Thanks @Gallaecio For the detailed help

spidermon/contrib/scrapy/monitors.py
rosheen33: @rennerocha If it is sent through command line as @Gallaecio mentioned above then it will be str

Gallaecio: Actually, once you load them with `json.loads`, they should be `int`.

spidermon/contrib/scrapy/monitors.py
rosheen33: Documentation verified.

spidermon/contrib/scrapy/monitors.py
rosheen33: -----> removed.

spidermon/contrib/scrapy/monitors.py
rosheen33: Done

spidermon/contrib/scrapy/monitors.py
rosheen33: @Gallaecio What if it is a list and we split the value based on 
``return value.split(',')``
Would we need that int(code) then ?

Gallaecio: You are completely right, I had forgotten about comma-separated lists being supported as well!

rosheen33:
@rennerocha I have changed the code according to your comments.
Please let me know if its alright and changes needed further ?

rosheen33:
@raphapassini Please have a look

PR #159, opened 2019-05-13
Slackclient version problem

### LINKED ISSUE #158: Spidermon doesn't work with slackclient version 2.x
Action to send [Slack notification](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/slack/__init__.py#L5) doesn't work with [slackclient](https://pypi.org/project/slackclient/) version 2.x .

The provided a [guide](https://github.com/slackapi/python-slackclient#migrating-from-v1) for migration (that only works with Python 3.6 or higher).

rennerocha:
This should fix #158 . As slackclient version 1.3.* still works and the actual version don't work with Python lesser than 3.6, I didn't update the code to use newest version of the library. When spidermon become Python3.6+ only or the library stop working, we will need to change our action.

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -165,6 +165,17 @@ SPIDERMON_EMAIL_SUBJECT_TEMPLATE
 Slack action
 ============
 
+.. warning::
+
+   You need to install `slackclient`_ to use this feature.
+
+   .. code-block:: shell
+
+     $ pip install "slackclient>=1.3,<2.0"
+
+   **Do not install** a version greater than **1.3.*** because newer versions
+   work only with Python 3.6+ and contain incompatible features.
+
 This action allows you to send custom messages to a `Slack`_ channel (or user)
 using a bot when your monitor suites finishes their execution. To use this action
 you need to provide the `Slack credentials`_ in your `settings.py`
@@ -538,3 +549,5 @@ the `run_action` method.
         def run_action(self):
             # Include here the logic of your action
             # (...)
+
+.. _`slackclient`: https://pypi.org/project/slackclient/
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -165,20 +165,21 @@ SPIDERMON_EMAIL_SUBJECT_TEMPLATE
 Slack action
 ============
 
-.. warning::
-
-   To use this feature you need to install `slackclient`_ **1.3 or higher, but
-   lower than 2.0**. Version 2.0 is Python3.6+ only and contains incompatible
-   features.
-
-   .. code-block:: shell
-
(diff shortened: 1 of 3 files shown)


docs/source/actions.rst
```diff
@@ -165,6 +165,17 @@ SPIDERMON_EMAIL_SUBJECT_TEMPLATE
...
 
+.. warning::
+
+   You need to install `slackclient`_ to use this feature.
+
+   .. code-block:: shell
+
+     $ pip install "slackclient>=1.3,<2.0"
+
+   **Do not install** a version greater than **1.3.*** because newer versions
```
manycoding: There could be maintenance releases before 2 which will still work.
Apart from this everything looks good.

docs/source/actions.rst
```diff
@@ -165,6 +165,17 @@ SPIDERMON_EMAIL_SUBJECT_TEMPLATE
 Slack action
 ============
 
+.. warning::
```
Gallaecio: Instead of making this a warning, I think it may be better, documentation-wise, to refactor the initial paragraph instead.

For example:

```
This action allows you to send custom messages to a `Slack`_ channel (or user)
using a bot when your monitor suites finish their execution.

To use this action you need to:

#.  Install `slackclient`_ 1.3 or higher, but lower than 2.0:

    .. code-block:: shell

        $ pip install "slackclient>=1.3,<2.0"

#.  Provide the `Slack credentials`_ in your ``settings.py`` file as follows:
```

rennerocha: I chose a warning instead of plain-text because I really want the user to notice as fast as possible that she/he needs to install a specific version of slackclient. If we keep this information in a regular text, some user may not notice that quickly, and this has a potential to create more issues in future. What do you think?

BTW, I liked the sentence you included. I changed it in the warning using part of it.

Gallaecio: Still, the installation of `slackclient` should be part of the setup instructions. And I don’t think sections should start with warnings, I think warnings should appear right after the point they warn about; right now you are warning about a potential dependency version issue before you’ve explained what the action does.

What about putting the warning after the corresponding installation step? e.g.

```
This action allows you to send custom messages to a `Slack`_ channel (or user)
using a bot when your monitor suites finish their execution.

To use this action you need to:

#.  Install `slackclient`_ 1.3 or higher, but lower than 2.0:

    .. code-block:: shell

        $ pip install "slackclient>=1.3,<2.0"

    .. warning:: This action does not work with `slackclient`_ 2.0 or later.

#.  Provide the `Slack credentials`_ in your ``settings.py`` file as follows:
```

rennerocha: Much better!

Issue #151, opened 2019-04-19
Per-field required percentage checking


wRAR:
Right now Spidermon field percentage validation works in this manner: 

- you set some of your fields as required in the schema
- the validation code counts all missing required fields in all items, by catching schema validation errors
- the error count is divided by the item count

The result is not very useful as it's not clear what threshold to set because all missing fields in all items are combined into one number.

What we actually need is a percentage check for each required field, or, better, separate percentages for all fields. How do I see this:

- _something_ calculates the field count for each field
- `ValidationInfo` calculates percentages for each field by dividing counts by the item count
- `ValidationMonitorMixin` gets per-field, per-spider and per-project thresholds and compares the percentage for each field with the relevant threshold
- an error, or a bunch of errors, is emitted, listing all failing fields with their actual and expected percentages

Some implementations I've seen:

- count items in a pipeline, store counts in the Spider object; get the thresholds from settings and spider attributes
- put thresholds in the JSON schema for required fields and retrieve them in the monitor; compare per-field validation errors (which spidermon already counts) with the thresholds

The main question here is where to count the fields and where to store them. `ItemValidationPipeline` seems to be the ideal place for the counting code as it already processes all items. The data can be stored in the stats, though I'm not sure what is the policy of adding multiple new keys (one per field) to stats. If this is a problem the code can be disabled unless some setting is set.

The thresholds can be taken from settings and spider attributes, it should be possible to pass a single number or a dict of field: number mappings, I don't think this is a problem.

To keep the compatibility the old methods should be kept and new ones created, though I'm not sure if the old ones are useful as is.

We can also want to check percentages for subfields, it's possible with the current code as jsonschema emits errors for them too, though because of mangling during converting errors to stats keys some guesswork is needed to unmangle them. My current proposal doesn't take this additional feature into account, suggestions welcome.

muzaffaryousaf:
@wRAR Can we change the existing methods and mechanism in such way that if additional/optional fields are provided then we do validation in the suggested way rather than having duplicated code with minor differences ?

ejulio:
The idea looks good. My thoughts are:

* Can't we use the stats already provided by `ItemValidationPipeline`? It stores missing fields as _spidermon/validation/fields/errors/missing_required_field/field_name_
* The thresholds should be settings, so that we can set them in settings.py or thought `custom_settings`.
* Probably, they should be `dict`s or a number (default for all fields)
[...]

Issue #144, opened 2019-04-09
Validation monitor only raises error for first invalid field at a time


HassanQamar07:
The validation monitor `check_missing_required_fields`, `check_missing_required_fields_percent`, `check_fields_errors` and `check_fields_errors_percent` monitors only raises error for the first effected field because the related methods `_check_missing_required_count`, `_check_missing_required_percent`, `_check_field_errors` and `_check_field_errors_percent` is called in loop for each field and it raises error for effected field and breaks the loop for remaning fields.

https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/monitors/mixins/validation.py#L93

muzaffaryousaf:
@wRAR Please see the description of this issue.

muzaffaryousaf:
@wRAR  Will this be addressed under #164  as well ?

wRAR:
@muzaffaryousaf yes.

wRAR:
This is implemented in #164, now all errors are raised together.

PR #164, opened 2019-05-29
feature: Allow per-field checking in ValidationMonitorMixin

wRAR:
This is the first step for #151 and is a partial fix for #144. Note the code duplication. I don't see any good ways to do this because of the current code structure of short methods calling each other and ending with an assert. It seems the only good way here is `subTest` from Python 3.4 or `unittest2`. You may consider this as a call for a discussion.

It also seems there is no tests for this code.


spidermon/contrib/monitors/mixins/validation.py
```diff
@@ -121,11 +121,24 @@ def check_missing_required_fields_percent(
...
-            for field_name in field_names:
-                self.check_missing_required_field_percent(field_name, allowed_percent)
+            field_names = self.validation.fields
+        msgs = []
+        for field_name in field_names:
+            missing_percent = (
+                self.validation.fields[field_name].errors["missing_required_field"].percent
+            )
+            if missing_percent > allowed_percent:
+                msg = "{percent}% of required field {field} are missing!{threshold_info}".format(
```
ejulio: At least this message could be extracted in a method, since it is duplicated a couple of lines below.
Not completely sure if it would be helpful, but, maybe, we can add readability to the code.
Mostly because the method names are quite similar and looking, for the first time, through the file I got lost.

wRAR: Yeah, I don't like these methods, I would like them to be improved too.

wRAR:
@ejulio as there seem to be no tests for this file, where do you propose to put them?

ejulio:
@wRAR I'd create them under `tests/` or `tests/contrib`.
Usually I prefer to follow the same path for source and tests, but tests for built-in monitors are under`tests/` and their implementation is under `contrib/scrapy/..`.

Maybe @rennerocha has a better idea here.

rennerocha:
> @wRAR I'd create them under `tests/` or `tests/contrib`.
> Usually I prefer to follow the same path for source and tests, but tests for built-in monitors are under`tests/` and their implementation is under `contrib/scrapy/..`.
> 
> Maybe @rennerocha has a better idea here.

I agree with you. Following the same paths for source and tests is a good practice making it easier to find the related tests.

In this case, put the tests inside `tests/contrib/monitors/mixins/` seems a good idea.

wRAR:
@ejulio @rennerocha I've rewritten all of the methods in `ValidationMonitorMixin` to have the same behavior, not just `check_missing_required_fields_percent`:

* single-check methods work as they did, checking one field and raising an error - no problem here.
* multiple-check methods, when the `field_names` list is passed, check all passed fields individually and combine the errors into one, instead of raising the first one - this is an improvement (fixes #144)  and it shouldn't break any assumptions.
* multiple-check methods, when the `field_names` list is not passed, check all fields with errors (`self.validation.fields`) individually just like the previous case, instead of doing weird aggregation of errors - this fixes #151 but is backward incompatible.

I don't know how to keep the backward compatibility, except for adding some flag to all such methods, which enables the new behavior, what do you guys think?

ejulio:
Since they follow the same API in terms of input, the output is barely the same, just a matter of aggregations. I'd go with the strategy pattern.
Something like

```
class ValidationMonitorMixin:
    def __init__(self, break_on_first_error):
        if break_on_first_error:
            self._validation_mixin = ValidationMonitorMixinBreakFirstError()
        else:
            self._validation_mixin = ValidationMonitorMixinAggregateErrors()

   def check_missing_required_field_percent():
       return self._validarion_mixin.check_missing_required_field_percent()

class ValidationMonitorMixinBreakFirstError():
   # all the logic to break in the first error
    pass

class ValidationMonitorMixinAggregateErrors():
   # all logic to aggregate errors
    pass
```

What do you think?

wRAR:
@ejulio I think not breaking on the first error is not something that we want to keep, I'm concerned about the old numbers aggregation, like counting all field errors together instead of per-field. On the other hand, your idea can be applied to this just as well.

So one option is having the old class completely unchanged and a separate new class with new code.

Another option would be adding a flag to all affected methods. Here is a sample of the refactored code with the new behavior which has the old code inserted into it:

```python
def check_missing_required_fields(self, field_names=None, allowed_count=0, new_behavior=False):
    """
    Checks that the number of "missing_required_field" errors for the ``field_names`` fields is less or equal than
    ``allowed_count`` and raises an error with all problematic fields.
    If ``field_names`` is None and ``new_behavior`` is True, checks all fields.
    If ``field_names`` is None and ``new_behavior`` is False, checks that the total number of
    "missing_required_field" errors is less or equal than ``allowed_count``.
    """
    if not new_behavior and not field_names:
        # TODO: deprecate this
        missing_count = self.validation.errors["missing_required_field"].count
        self.assertLessEqual(
            missing_count,
            allowed_count,
            msg="{count} required field{plural} are missing!{threshold_info}".format(
                count=missing_count,
                plural="" if missing_count == 1 else "s",
                threshold_info=(" (maximum allowed %d)" % allowed_count)
                if allowed_count > 0
                else "",
            ),
        )
        return
    if not field_names:
        field_names = self.validation.fields
    msgs = []
    for field_name in field_names:
        missing_count = self.validation.fields[field_name].errors["missing_required_field"].count
        if missing_count > allowed_count:
            msg = self._get_msg_for_missing_required_count(field_name, missing_count, allowed_count)
            msgs.append(msg)
    if msgs:
        msgs.insert(0, 'Required fields are missing:')
        self.fail('\n'.join(msgs))

def check_missing_required_field(self, field_name, allowed_count=0):
    """
    Checks that the number of "missing_required_field" errors for the ``field_name`` field is less or equal than
    ``allowed_count``.
    """
    missing_count = (
        self.validation.fields[field_name].errors["missing_required_field"].count
    )
    msg = self._get_msg_for_missing_required_count(field_name, missing_count, allowed_count)
    self.assertLessEqual(missing_count, allowed_count, msg)

@staticmethod
def _get_msg_for_missing_required_count(missing_count, allowed_count, field_name=None):
    msg = "Required field {field} is missing in {count} items!{threshold_info}".format(
        count=missing_count,
        field=field_name,
        threshold_info=(" (maximum allowed %d)" % allowed_count)
        if allowed_count > 0
        else "",
    )
    return msg
```

I didn't want to spend time combining the old and the new message into one format so I inserted it as is, and it looks like it's not worth it to do a separate class with just one relatively small change. I like the idea of putting the behavior switch flag into the class instead of each method though.

So, should we just keep the old class side by side or combine them?

ejulio:
If it makes things easier, then I think we can keep it.
Otherwise, my concern is that we'll need to add this flag to all methods.
So, maybe keep it only in `__init__` and create separate methods instead of classes.
My idea is that, by separating these concerns it might be easier to understand, maintain and test..
Also, it might be an idea to get rid of those methods with similar names..

wRAR:
@ejulio I've implemented that and wrote tests for all methods for both old and new behavior.

My only concern about setting the flag in `__init__`: as this is a mixin I'm not sure how should this be used in user code. In the test I just set the variable instead of calling mixin's `__init__`.

spidermon/contrib/monitors/mixins/validation.py
```diff
@@ -84,125 +84,259 @@ def __init__(self, stats, prefix=None):
 
 
 class ValidationMonitorMixin(StatsMonitorMixin):
+    def __init__(self, new_behavior=False):
+        super(ValidationMonitorMixin, self).__init__()
+        self.new_behavior = new_behavior  # how to handle empty field lists
```
ejulio: Maybe a better name here would avoid the comment and make the `if` statements below more readable

tests/contrib/monitors/mixins/test_validation.py
```diff
@@ -0,0 +1,214 @@
...
+class DummyValidationMonitor(BaseScrapyMonitor, ValidationMonitorMixin):
+    def __init__(self, stats, methodName="runTest", name=None):
+        super(DummyValidationMonitor, self).__init__(methodName, name)
+        self.data = Data({'stats': stats})
+        self.new_behavior = True
+
+    def runTest(self):
+        pass
+
+
```
ejulio: I think that at least one test with `allowed_count > 0` to check the threshold message would be nice

tests/contrib/monitors/mixins/test_validation.py
```diff
@@ -0,0 +1,214 @@
...
+    'spidermon/validation/fields/errors/': 10,
+    'spidermon/validation/items': 10,
+}
+
+
+class DummyValidationMonitor(BaseScrapyMonitor, ValidationMonitorMixin):
+    def __init__(self, stats, methodName="runTest", name=None):
+        super(DummyValidationMonitor, self).__init__(methodName, name)
+        self.data = Data({'stats': stats})
+        self.new_behavior = True
```
ejulio: This is related to your comment, so it wold be nice to have a check for `new_behavior = False`.
Maybe, creating pytest fixtures for `new_behavior = False` and `new_behavior = True`.
It looks like we can call `__init__` in the mixin https://stackoverflow.com/questions/6098970/are-mixin-class-init-functions-not-automatically-called

spidermon/contrib/monitors/mixins/validation.py
```diff
@@ -84,125 +84,259 @@ def __init__(self, stats, prefix=None):
...
+        "missing_required_field" errors is less or equal than ``allowed_count``.
+        """
+        if not self.new_behavior and not field_names:
+            # TODO: deprecate this
             missing_percent = self.validation.errors["missing_required_field"].percent
-            self._check_missing_required_percent(missing_percent, allowed_percent)
-        else:
-            for field_name in field_names:
-                self.check_missing_required_field_percent(field_name, allowed_percent)
+            self.assertLessEqual(
```
ejulio: Here is an assertion for _less or equal_, but a couple of lines below, it is for _greater than_.
The behavior is the same, but keeping only one conditional would make it easier to understand.

spidermon/contrib/monitors/mixins/validation.py
```diff
@@ -84,125 +84,259 @@ def __init__(self, stats, prefix=None):
...
+                    if allowed_count > 0
+                    else "",
+                ),
+            )
+            return
+        if not field_names:
+            field_names = self._get_all_fields()
+        msgs = []
+        for field_name in field_names:
+            errors_count = self._get_errors_count(errors, field_name)
```
ejulio: I think we are missing an `if` statement here.
Otherwise, it will add a message even if there is no error for the field.
If this case holds, it would be nice to add a test case

spidermon/contrib/monitors/mixins/validation.py
```diff
@@ -84,125 +84,259 @@ def __init__(self, stats, prefix=None):
...
+                    if allowed_percent > 0
+                    else "",
+                ),
+            )
+            return
+        if not field_names:
+            field_names = self._get_all_fields()
+        msgs = []
+        for field_name in field_names:
+            errors_percent = self._get_errors_percent(errors, field_name)
```
ejulio: Same as above here, maybe an `if` statement is missing

spidermon/contrib/monitors/mixins/validation.py
```diff
@@ -84,125 +84,259 @@ def __init__(self, stats, prefix=None):
...
+        equal than ``allowed_count``.
+        """
+        if not self.new_behavior and not field_names:
+            # TODO: deprecate this
             errors_count = self.validation.errors.count
-            self._check_field_errors(errors_count, allowed_count)
-        else:
-            for field_name in field_names:
-                self.check_field_errors(field_name, errors, allowed_count)
+            self.assertLessEqual(
```
ejulio: All the methods look quite similar, this first `if` statement then a `for` loop.
Maybe, if you change this assertion to an `if` statement followed by `self.fail` (as below) it would give some insight on how to extract this behavior in a method.
Then, these functions could be only a first check and preparing the proper inputs checks for this method.
More of an idea for refactoring that can be done later :smile:

tests/contrib/monitors/mixins/test_validation.py
wRAR: _old() tests set new_behavior = False and test it, but I agree it can be useful to have them more separate.

spidermon/contrib/monitors/mixins/validation.py
wRAR: The assert line means "it's good if it's <=" while the manual comparison line means "it's bad if it's >", so just reversing the manual comparison to read `if allowed_percent < missing_percent` looks unhelpful to me, should it read `if not (missing_percent <= allowed_percent)`?

ejulio: It is more of an idea.
If you think it is understandable as is, it is fine.
My concern was only to keep the statements similar.

wRAR:
@ejulio I've resolved the problems you've found and added some additional tests, I think refactoring is a good idea but should be done after the improvements are merged. Let me know if you think anything should be improved further.

tests/contrib/monitors/mixins/test_validation.py
```diff
@@ -199,16 +221,15 @@ def test_check_fields_errors_percent_multiple_fields():
         monitor.check_fields_errors_percent(field_names=['field2', 'field3'])
 
 
-def test_check_fields_errors_percent_no_fields_old():
-    monitor = DummyValidationMonitor(stats)
-    monitor.new_behavior = False
+def test_check_fields_errors_percent_no_fields_old(old_monitor):
+    monitor.correct_field_list_handling = False
```
ejulio: I think it should be `old_monitor`, right?

wRAR: It's just an unused line :) I'll remove it.

rennerocha:
@wRAR code looks good and certainly it is much more useful now!!! :-) I see some code duplication, but I believe that we can try refactor it later.

Some small comments:
1. Maybe we should add some DeprecationWarning in the old behaviour code, as probably we will remove it in next versions and it is good to warn potential new users about it (and we remove the #TODOs comments!)
1. I miss documentation about ValidationMixin. As it is a very useful feature, we should add it before merging this update.

wRAR:
@rennerocha I've implemented both items, please check.

PR #165, opened 2019-05-31
Improving docs

ejulio:
We had an issue when building the docs using `make html` because it wasn't creating autodocs from docstrings.

I updated `conf.py` to include a path to spidermon.
Also updated the contributing guide adding the commands to build the docs.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -90,13 +90,20 @@ including testing other Python versions with tox::
 
   $ black .
 
-7. Commit your changes and push your branch to GitHub::
+7. [Optional] If you changed something related to docs, make sure it compiles properly
+
+    $ cd docs
+    $ make html
+
+    Now you can go to `spidermon/docs/build` and open `index.html`
+
+8. Commit your changes and push your branch to GitHub::
 
     $ git add .
     $ git commit -m "Your detailed description of your changes."
     $ git push origin name-of-your-bugfix-or-feature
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.rst
```diff
@@ -90,13 +90,20 @@ including testing other Python versions with tox::
 
   $ black .
 
-7. Commit your changes and push your branch to GitHub::
+7. [Optional] If you changed something related to docs, make sure it compiles properly
```
rennerocha: Compile the docs is not optional. You need to make sure that the docs were generated correctly before sending the PR. Also includes the need to install requirements-docs.txt so it is possible to compile it.

> If you changed something related to docs, **make sure it compiles properly**:

manycoding: @ejulio Do you want to include this compile as a CI check here so if it's failed, it's visible?

Gallaecio: Otherwise, maybe we could add the docs to CI in a different pull request. Scrapy could serve as a starting point here, documentation warnings are treated as errors by the CI there.

rennerocha: It is a good idea to include the build in CI. I created #169 to track this separately from this PR.
Anyway, add to the instructions that build the docs is not an optional step is needed.

ejulio: hey @rennerocha , updated :)

PR #167, opened 2019-06-04
Fixed AttributeError when obtaining location of the stats file for a …

rennerocha:
…spider


rennerocha:
> The code looks good to me, but could you please point out the issue link or explain how you got the error? Just for me to better understand

Is is a bug from the code used to solve #122

Issue #126, opened 2019-03-14
How to set a local/custom template in Email actions


matiskay:
Hi,

I would like to know if there is way to pass  a local jinja template to the actions. I'm setting a local path and absolute path in `SPIDERMON_BODY_HTML_TEMPLATE`  but  I'm getting a `TemplateNotFound` error.

I tried `SPIDERMON_BODY_HTML_TEMPLATE = 'reports/email/monitors/result.jinja'` and it worked but I would like to add my own template.

It looks like jinja2 loader only looks at `/spidermon/contrib/actions/reports/templates/` folder inside `spidermon` package.

rennerocha:
Try creating a custom action and include the path for your template inside `template_paths` [class attribute](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/templates.py#L27). This should solve your problem for now, but I believe it would be a good idea to add a new setting to allow us to define a list of locations that we want the templates to be searched against.

rennerocha:
When trying to [load a template](https://github.com/scrapinghub/spidermon/blob/master/spidermon/templates.py#L74) that is an absolute path, Jinja2 loader didn't look at the right place, but instead it concatenates the provided absolute path to the paths loaded in the `Environment`.

At this moment you can not use absolute paths for Spidermon's custom templates. You need to create a custom action inherited from `[ActionWithTemplates](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/templates.py#L26)` and define the `[template_paths](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/templates.py#L27) containing the absolute location of your desired template.

rennerocha:
Considering that my template is located in `/absolute/location/of/my/template/template_name.jinja`

```
from spidermon.contrib.actions.email import SendEmail

class MyCustomSendEmail(SendEmail):
  template_paths = ["/absolute/location/of/my/template"]
```

and add to `settings.py`:

```
SPIDERMON_BODY_HTML_TEMPLATE = "template_name.jinja"
````

rennerocha:
@matiskay Spidermon 1.10.2 allows you to set an absolute path for your custom templates.
If you find any problems, refer to this issue again.

PR #170, opened 2019-06-06
bug: Allowed to add absolute location for custom templates

rennerocha:
This should address the problem in #126


spidermon/templates.py
```diff
@@ -71,6 +71,11 @@ def reload_env(self):
             self.env.globals[global_name] = global_value
 
     def get_template(self, name):
+        if os.path.isabs(name):
+            self.paths.append(os.path.dirname(name))
```
ejulio: why not `self.add_path`?

rennerocha:
> Usually, I don't expect a _query method_ (`get`) to change the state of my object.

It makes sense. `get_template` should never update the content of `paths`.
The problem we have is that when you include an absolute path for a template, it will never be found because when the TemplateLoader is loaded, the path of the absolute template is not added.

I believe the right solution would be update the path in `auto_discover` or `auto_discover` methods of the template loader, using the info from the settings provided.

spidermon/templates.py
rennerocha: Because I didn't notice that it already exists :-) I will change that as well.

rennerocha:
> I'm not familiar with this part of spidermon, so I'll just leave the comment 
> But, to me, it looks like a side effect that shouldn't be there.
> I should add the path somehow, or it should be clear that the path can be added to the path list.
> Usually, I don't expect a _query method_ (`get`) to change the state of my object.

To update the paths in a different place, I need to know beforehand the name of all settings of actions that are related to a template, and it won't work if we are dealing with custom actions.

Other options I can see:

- Create a new setting with the path where to look templates - it is something interesting to be added, but still will fail with absolute paths in template settings;
- Instead of update the template_loader path, create a temporary `Environment` just to return the absolute file template when exists, and then get rid of the side effect and the `get_template` method keeps doing its job (I will try this one).

spidermon/templates.py
```diff
@@ -62,16 +76,17 @@ def discover_folder(self, candidate_folder):
...
-            self.env.filters[filter_name] = filter
-
-        for global_name, global_value in GLOBALS.items():
-            self.env.globals[global_name] = global_value
+        self.env = get_environment(self.paths)
 
     def get_template(self, name):
-        return self.env.get_template(name)
+        template = None
+        try:
```
ejulio: No need to work here, but we could avoid the exception just by checking if it is an absolute path.
If it is an absolute path, then load with your solution.
Otherwise, keep as before this change.
The current approach, from what I can think, might fail in a scenario (not sure if it is expected, though):
Say that I provided a template name that does not exists, should it raise `TemplateNotFound` or `return None`?
If it should raise the exception, then it will fail in the current approach.
Then, if the idea above holds, it would be nice to add a test case to ensure the expected scenario.

PR #171, opened 2019-06-10
feature: Added default Value for 'SPIDERMON_BODY_HTML_TEMPLATE' setting

rosheen33:
SPIDERMON_BODY_HTML_TEMPLATE = 'reports/email/monitors/result.jinja'


rosheen33:
@raphapassini I think we should make it generic.

rosheen33:
@raphapassini  Can't we just disable the buttons with links and that will be it ?

rosheen33:
@raphapassini These SC links will only be enabled if 
we pass the context will that not work ?
```
 SPIDERMON_EMAIL_CONTEXT: {"show_job_button": True} 
```

rosheen33:
@rennerocha Can you please also review ?

docs/source/actions.rst
```diff
@@ -102,6 +102,12 @@ Default: ``None``
 SPIDERMON_BODY_HTML_TEMPLATE
 ----------------------------
 
+String containing the location of the template for the spidermon email report.
```
Gallaecio: :lipstick: 

```suggestion
String containing the location of the template for the Spidermon email report.
```

Also, I think it would be great to have here a link to a topic that covers how to write such a template. Since at this point there is no such a topic, I think that we could say ‘Jinja2 template’ instead of just ‘template’, and link ‘Jinja2’ to the Jinja2 documentation.

docs/source/actions.rst
```diff
@@ -102,6 +102,12 @@ Default: ``None``
...
+
+Default: ``reports/email/monitors/result.jinja``
+
+Link to the Default `Template`_.
```
Gallaecio: I would not link to the `master` template, because the documentation should always be specific to the version of Spidermon containing the documentation.

Maybe you could remove this line and use [`:download:`](https://stackoverflow.com/a/8622242/939364) on ``reports/email/monitors/result.jinja`` above.

spidermon/contrib/actions/email/__init__.py
```diff
@@ -22,7 +22,8 @@ class SendEmail(ActionWithTemplates):
     body_text = None
     body_text_template = None
     body_html = None
-    body_html_template = None
+    # Default value for template
```
Gallaecio: :lipstick: I’m not sure the comment is really necessary.

rosheen33: Alright I will remove the comment :)

rosheen33:
@Gallaecio Please have a look again
I have changed to conform with your comments.

PR #183, opened 2019-06-27
bug: Correctly send slack message to logs when fake setting is enabled

rennerocha:
When `SPIDERMON_SLACK_FAKE = True`, the content of the message is not sent correctly to the logs.

```
======================================================================
ERROR: SendSlackMessageSpiderFinished
  File "/home/renne/projects/scrapinghub/spidermon/spidermon/core/actions.py", line 37, in run
    self.run_action()
  File "/home/renne/projects/scrapinghub/spidermon/spidermon/contrib/actions/slack/__init__.py", line 204, in run_action
    logger.info(message.as_string())
AttributeError: 'str' object has no attribute 'as_string'
```


raphapassini:
Nice bug fix, maybe worth testing that everything is actually working as expected?

rennerocha:
@raphapassini added some tests for the logging feature. Could you please review it?

Issue #189, opened 2019-07-17
Use other regions on AWS SES


thamenato:
Hi, I currently ran into that problem and couldnt really solve it by simply adding the environment variable `AWS_DEFAULT_REGION` into something else than `us-east-1`. I made a local patch to use a different function from `boto.ses` which allows you to set the `region_name`.

I would like to know if I can send a pull request and how I can do it. 

Thanks

vipulgupta2048:
Hi @thamenato, thanks for using Spidermon. Could you tell a little more about the exact problem you faced that would be helpful to debug this on our end? Regarding the PR, yes, you can send a PR. One of the maintainers can review and suggest changes to it. 

Guide to making a PR --> https://help.github.com/en/articles/creating-a-pull-request-from-a-fork

thamenato:
@vipulgupta2048  Just sent it! https://github.com/scrapinghub/spidermon/pull/190

My problem was that my SES is configured on `us-west-2` and when the class SendSESEmail tries to send the email with the function `session.send_raw_email()` it raises the following exception:
```
boto.ses.exceptions.SESAddressNotVerifiedError: SESAddressNotVerifiedError: 400 Email address is not verified.
...
... Email address is not verified. The following identities failed the check in region US-EAST-1: ...
```
So I figured that the problem was that boto was ignoring (or not even using) my `DEFAULT_AWS_REGION` environment variable.

Once I changed the function and set the region it worked like a charm.

rennerocha:
Hello @thamenato ,

I believe that you need to change [this method](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/email/ses.py#L33) so the region can be set. This would be a great improvement in our SES e-mail action.

Feel free to send a PR so we can add this new option in our next Spidermon release.

PR #190, opened 2019-07-17
feature: Add option to set AWS Region Name on SES E-Mail action

thamenato:
The function `boto.connect_ses` does not allow the usage of a different region other than the default `us-east-1`.
With this commit it will be possible to set the AWS region by adding `SPIDERMON_AWS_REGION` to the settings.py.

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -70,8 +70,8 @@ AWS Secret Key.
 
 Default: ``None``
 
-SPIDERMON_AWS_REGION
-------------------------
+SPIDERMON_AWS_REGION_NAME
+-------------------------
 
 AWS Region.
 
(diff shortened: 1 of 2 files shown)


docs/source/actions.rst
```diff
@@ -70,6 +70,13 @@ AWS Secret Key.
 
 Default: ``None``
 
+SPIDERMON_AWS_REGION
```
rennerocha: Scrapy uses [AWS_REGION_NAME](https://docs.scrapy.org/en/latest/topics/settings.html#aws-region-name) as setting name. I would prefer if we try to keep a similar name. `SPIDERMON_AWS_REGION_NAME` is a better name.

thamenato: Absolutely! Done the change here.

rennerocha:
Solves #189

PR #204, opened 2019-08-22
feature: add-smtp-mail

sulthonzh:
hello i just created new PR for smtp mail action

### DIFF
--- a/spidermon/contrib/actions/email/smtp.py
+++ b/spidermon/contrib/actions/email/smtp.py
@@ -1,0 +1,62 @@
+from __future__ import absolute_import
+import smtplib
+
+from spidermon.exceptions import NotConfigured
+
+from . import SendEmail
+
+
+class SendSmtpEmail(SendEmail):
+    smtp_host = "smtp.gmail.com"
+    smtp_port = 587
+    smtp_user = None
+    smtp_password = None
+    
+    def __init__(
+        self,
+        smtp_host=None,
+        smtp_port=None,
+        smtp_user=None,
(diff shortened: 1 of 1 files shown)


spidermon/contrib/actions/email/smtp.py
```diff
@@ -0,0 +1,62 @@
...
+    def send_message(self, message):       
+        if not self.sender:
+            raise NotConfigured("You must provide the sender.")
+        if not self.to:
+            raise NotConfigured("You must provide the receiver.")
+
+        if not self.sender or not self.to:
+            return
+
+        server = smtplib.SMTP(self.smtp_host, self.smtp_port)
```
rennerocha: Sometimes using [smtplib.SMTP](https://docs.python.org/2/library/smtplib.html#smtplib.SMTP) is not appropriated and you need to use [smtplib.SMTP_SSL](https://docs.python.org/2/library/smtplib.html#smtplib.SMTP_SSL). It will be interesting to have an option so the user can define which one to use depending on the needs of the SMTP server being connected.

ejulio: Maybe it is worth checking https://docs.scrapy.org/en/latest/topics/email.html

rennerocha: It makes sense. Using the built-in SMTP method from Scrapy would be a good solution.

rennerocha:
This feature is being continued in https://github.com/scrapinghub/spidermon/pull/345

PR #205, opened 2019-08-23
feature: CLI implementation

leovictorsr:
Part of my GSoC project.
This implements the CLI and the commands `version` that shows Spidermon's installed version, and `setup`, that enables Spidermon if it's not enabled, and the monitors of the Basic Scrapy Suite with its settings.


rennerocha:
We don't need a CLI for Spidermon at this moment, so we will close this PR.

PR #212, opened 2019-09-16
feature: Add fallback to scrapy AWS settings

### LINKED ISSUE #211: Fallback to scrapy AWS settings if not set
Currently, to send _SES_ emails, we need to set `SPIDERMON_AWS_ACCESS_KEY_ID` and the other configs for that.
Though, it may be the case of a single user configured in scrapy for file delivery and the same user is authorized to send emails.
My idea is to read `SPIDERMON_AWS_ACCESS_KEY_ID`, and if it is empty, we can fallback to `scrapy` default [`AWS_ACCESS_KEY_ID`](https://docs.scrapy.org/en/latest/topics/feed-exports.html#s3).
This way, if a single user is authorized to perform all the actions, we only need to set it once, not replicate the same configs.

ejulio:
Fix #211 
It is quite common to have the same user for data delivery in S3, URL input and notification.
If we have all of these cases, we need to set the same info in three different settings.
This PR adds a fallback from spidermon settings to scrapy ones.
Spidermon settings didn't follow scrapy names, so I added the proper ones and a deprecation warning for spidermon ones.


spidermon/utils/settings.py
```diff
@@ -15,3 +16,26 @@ def getdictorlist(crawler, name, default=None):
...
+
+def get_aws_credentials(settings):
+    aws_access_key_id = settings.get("SPIDERMON_AWS_ACCESS_KEY")
+    aws_secret_access_key = settings.get("SPIDERMON_AWS_SECRET_KEY")
+
+    if aws_access_key_id and aws_secret_access_key:
+        warnings.warn(
+            "SPIDERMON_AWS_ACCESS_KEY and SPIDERMON_AWS_SECRET_KEY are deprecated. "
+            "Please update them to SPIDERMON_AWS_ACCESS_KEY_ID and SPIDERMON_AWS_SECRET_ACCESS_KEY. "  # noqa
+            "Scrapy settings AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are also valid.",
```
ejulio: Particularly, I think these messages are a bit hidden.
We should have a stat in scrapy for counting it.
Or log it as an error after a given date (the max deprecation date).
Also, what do you think of adding a max date to this one, so we are sure to come back and remove it by the time.
Otherwise, it can stay here forever...

tests/utils/test_settings.py
```diff
@@ -0,0 +1,86 @@
+from scrapy.settings import Settings
```
rennerocha: There is no test to verify if the DeprecationWarning is raised and logged.

docs/source/actions.rst
```diff
@@ -61,15 +61,41 @@ Default: ``None``
...
+    This setting has been deprecated in preference of ``SPIDERMON_AWS_SECRET_ACCESS_KEY``.
+
+.. SPIDERMON_AWS_ACCESS_KEY_ID:
+
+SPIDERMON_AWS_ACCESS_KEY_ID
+---------------------------
+
 Default: ``None``
 
+AWS Access Key. If not set, it defaults to `AWS_ACCESS_KEY_ID`_.
```
rennerocha: It is not clear in this line that the default value is related to a Scrapy settings (without following the link). I believe that this could be clearer here.

docs/source/actions.rst
```diff
@@ -61,15 +61,41 @@ Default: ``None``
...
+
+.. SPIDERMON_AWS_SECRET_ACCESS_KEY:
+
+SPIDERMON_AWS_SECRET_ACCESS_KEY
+-------------------------------
+
+Default: ``None``
+
+AWS Secret Key. If not set, it defaults to `AWS_SECRET_ACCESS_KEY`_.
```
rennerocha: See comment on SPIDERMON_AWS_ACCESS_KEY_ID.

docs/source/actions.rst
```diff
@@ -568,3 +594,5 @@ the `run_action` method.
 
 .. _`slackclient`: https://pypi.org/project/slackclient/
 .. _Jinja2: http://jinja.pocoo.org/
+.. _`AWS_ACCESS_KEY_ID`: https://docs.scrapy.org/en/latest/topics/settings.html#std:setting-AWS_ACCESS_KEY_ID
```
rennerocha: Do you mind order the references in alphabetical order? :lipstick:

spidermon/utils/settings.py
rennerocha: You mean there is no counting on deprecation warnings? We have a stat for warnings.
I don't think that this needs to have a date, we can keep this message until we decide to remove the settings from the code (maybe when we release a new major version with some backward incompatible feature).

spidermon/utils/settings.py
```diff
@@ -15,3 +16,26 @@ def getdictorlist(crawler, name, default=None):
...
+            DeprecationWarning,
+        )
+
+    if not aws_access_key_id and not aws_secret_access_key:
```
rennerocha: What if someone mix the settings names? `SPIDERMON_AWS_ACCESS_KEY` and `SPIDERMON_AWS_SECRET_ACCESS_KEY` for example? I believe it would be better if we handle each setting separately.

spidermon/utils/settings.py
ejulio: But it is a generic one for warnings, so it may be lost with other stuff.
This is more of concern that, we should have it somewhere, so users have an idea on when it will be removed.
This is something I see in projects like numpy, scikit-learn and pandas.
Usually they warn you with a version where it will be dropped.

tests/utils/test_settings.py
ejulio: I don't if we should test it.
Do we test log messages?
Of course this is a bit different, but I don't know if it is a valid one.

spidermon/utils/settings.py
ejulio: Actually, I prefer them in pairs.
This is makes things clearer.
The case you mentioned, you are mixing the old style with the new one, so you should prefer the new one to avoid problems in the future.
In case of `SPIDERMON_AWS_SECRET_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY` I don't think it would happen because for scrapy you need to set both `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID`.
So, I think we can keep some consistency and the error happens earlier (because it won't work) instead of lost warning message.

tests/utils/test_settings.py
rennerocha: The log is there for some reason, so we need to ensure that it is being logged as expected.

ejulio:
@rennerocha and @Gallaecio , can you check this one again? :smile:

spidermon/utils/settings.py
rennerocha: As we don't have a data for removal of this setting, I believe we can just keep this warning. When we decide to remove it, we can update this message and include the version where this settings will be dropped.

rennerocha:
@ejulio all code now needs to be formatted with `black` . Could you please update your PR ?

rennerocha:
Replaced by https://github.com/scrapinghub/spidermon/pull/237 that contains black code formatting that makes the build pass correctly.

Issue #50, opened 2018-10-25
Option to make validation errors ERRORs in job


andrewbaxter:
At the moment validation errors can be added to the items in the _validation key, but there doesn't seem to be any way to make it more obvious that some errors are incorrect.

raphapassini:
@andrewbaxter can you test this again with the branch master?
I'm pretty sure that the errors are being reported as `log.ERRORS` and thus going to show up in Scrapy Cloud.

Gallaecio:
@andrewbaxter @raphapassini @rennerocha I also believe that is the case with the latest version of spidermon. Does that mean #50 here and #47 can be closed?

andrewbaxter:
Yep, this looks good to me!  Thanks!

andrewbaxter:
I guess I'll go ahead and close it.

andrewbaxter:
Sorry @Gallaecio :( I was looking at the wrong logs...

Here's what my logs look like when I get a validation error currently:
```
2019-04-05 20:25:45 [scrapy.core.scraper] DEBUG: Scraped from <200 https://scrapinghub.com/>
{'v': 'hi', '_validation': defaultdict(<class 'list'>, {'v': ['Invalid int']})}
2019-04-05 20:25:45 [scrapy.core.engine] INFO: Closing spider (finished)
2019-04-05 20:25:45 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
{'downloader/request_bytes': 876,
 'downloader/request_count': 4,
 'downloader/request_method_count/GET': 4,
 'downloader/response_bytes': 18973,
 'downloader/response_count': 4,
 'downloader/response_status_count/200': 1,
 'downloader/response_status_count/301': 2,
 'downloader/response_status_count/404': 1,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2019, 4, 5, 11, 25, 45, 200465),
 'item_scraped_count': 1,
 'log_count/DEBUG': 5,
 'log_count/INFO': 9,
 'memusage/max': 64413696,
 'memusage/startup': 64413696,
 'response_received_count': 2,
 'robotstxt/request_count': 1,
 'robotstxt/response_count': 1,
 'robotstxt/response_status_count/404': 1,
 'scheduler/dequeued': 2,
 'scheduler/dequeued/memory': 2,
 'scheduler/enqueued': 2,
 'scheduler/enqueued/memory': 2,
 'spidermon/validation/fields': 1,
[...]

Issue #47, opened 2018-10-05
Option to stop spider on errors


andrewbaxter:
I think it would be nice to have validation errors treated like other spider errors - increment the error counter, optionally stop the spider with an error condition if over an error count.

I'd primarily like to stop the spider so I made that the title, but having both would be nice.

Gallaecio:
Closing as per https://github.com/scrapinghub/spidermon/issues/50#issuecomment-480240128

andrewbaxter:
Sorry @Gallaecio :sweat: I rushed and accidentally closed #50 - alright to reopen this?

raphapassini:
I think this is a very good candidate for the next release.
We can add a new built-in monitor to close the spider when a number of errors happen.

I would like to suggest a new settings:
A `SPIDERMON_CLOSESPIDER_BY_STATS` where the key is a stats key and the value is the maximum number of errors for that given error for example:

```
SPIDERMON_CLOSESPIDER_BY_STATS = {
    "spidermon/validation/": 10,
}
```

We can based our `close_spider `method on the one from the built-in memory usage extension: 
https://github.com/scrapy/scrapy/blob/master/scrapy/extensions/memusage.py#L77

rennerocha:
Maybe expressions monitors can be used to achieve this, and we can use the pattern described in  [periodic monitors documentation](https://spidermon.readthedocs.io/en/latest/monitors.html#periodic-monitors). I am not yet quite sure that a built-in monitor is the best solution, as we may want to close the spider based on different assertions.

rosheen33:
@rennerocha @raphapassini What about `SPIDERMON_CLOSESPIDER_BY_EXPRESSIONS_MONITOR` ?
In which we check periodically for errors like we do in periodic monitors and end spider if certain limit is reached.
And allow all the expressions that we allow in Expressions Monitor ?

rosheen33:
Maybe Something like this ?
```
class SpiderPeriodicMonitorSuit(MonitorSuite):
    monitors_m = {
        "tests": [
            {"expression": "stats['item_scraped_count'] < 30"},
        ]
    }
    monitors = [create_monitor_class_from_dict(monitors_m, monitor_class=ExpressionsMonitor)]

    def on_monitors_failed(self, result):
        self._crawler.stop()
```

```
SPIDERMON_PERIODIC_MONITORS = {
    'myproject.monitors.SpiderPeriodicMonitorSuit': 10,
}
```

[...]

PR #216, opened 2019-09-23
Close spider by stats [#47][#50]

rosheen33:
Just an initial version of the solution.

I have added a new setting in **spidermon** as proposed by @raphapassini above
```
SPIDERMON_CLOSESPIDER_BY_STATS = {
    "spidermon/validation/": 10,
}
```
**Note:** Documentation is not added so far

I have implemented the solution 
Please have a look 

#47 
#50

### DIFF
--- a/spidermon/contrib/scrapy/pipelines.py
+++ b/spidermon/contrib/scrapy/pipelines.py
@@ -83,6 +83,7 @@ class ItemValidationPipeline(object):
                 "SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS"
             ),
             errors_field=crawler.settings.get("SPIDERMON_VALIDATION_ERRORS_FIELD"),
+            close_spider_by_stats=crawler.settings.get('SPIDERMON_CLOSESPIDER_BY_STATS'),
         )
 
     @classmethod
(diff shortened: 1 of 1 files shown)


spidermon/contrib/scrapy/pipelines.py
```diff
@@ -80,7 +83,14 @@ def set_validators(loader, schema):
                 "SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS"
             ),
             errors_field=crawler.settings.get("SPIDERMON_VALIDATION_ERRORS_FIELD"),
+            close_spider_by_stats=crawler.settings.get('SPIDERMON_CLOSESPIDER_BY_STATS'),
         )
+        pipeline._set_crawler(crawler)
```
Gallaecio: Why not pass it as a parameter to `cls()`?

spidermon/contrib/scrapy/pipelines.py
```diff
@@ -168,3 +181,19 @@ def _add_error_stats(self, errors):
...
+        This method evaluate each stat value and if no of errors are greater then
+        max_errors_allowed. Spider will be closed gracefully.
+        """
+        for stat_name, max_errors_allowed in stats_dict.iteritems():
+            stat_current_value = self.stats.stats.get_value(stat_name, 0)
+
+            if stat_current_value > max_errors_allowed:
+                self.crawler.engine.close_spider(
+                    spider,
+                    'Spidermon: No of errors are greater then the expected no of errors {} > {}'.format(
```
Gallaecio: I think we should not call them errors (they could be warnings or something else entirely), and I would include the name of the specific stat that triggers the spider closing.

```suggestion
                    'Spidermon: The value of stat {} ({}) went above the configured limit ({})'.format(stat_name, stat_current_value, max_errors_allowed)
```

PR #218, opened 2019-09-25
docs: Fixed warnings in docs build

vipulgupta2048:
:hatching_chick: 
Each commit fixes one error as referenced in #214

### DIFF
--- a/docs/source/monitors.rst
+++ b/docs/source/monitors.rst
@@ -32,7 +32,6 @@ number of items were extracted and fails if it is less than the expected thresho
                 msg='Extracted less than {} items'.format(minimum_threshold)
             )
 
-.. autoclass:: Monitor(name=name, data=None, parent=None)
 
 A `Monitor` instance defines a monitor that includes your monitoring logic and
 has the following properties that can be used to help you implement your monitors:
@@ -91,7 +90,7 @@ Here is an example of how to configure a new monitor suite in your project:
 
   ``name`` suite name
 
-  ``monitors`` list of :class:`Monitor` that will be executed if this suite is enabled.
+  ``monitors`` list of :class:``spidermon.core.monitors.Monitor`` that will be executed if this suite is enabled.
 
   ``monitors_finished_actions`` list of action classes that will be executed when
   all monitors finished their execution.
@@ -234,3 +233,5 @@ These are some of the usual metrics used in the monitors:
 - the amount of items that don't contain a specific field or a set of fields
 
 - the amount of items with validation errors (missing required fields, incorrect format, values that don't match a specific regular expression, strings that are too long/short, numeric values that are too high/low, etc.)
+
+.. autoclass:: Monitor(name=name, data=None, parent=None)
--- a/docs/source/monitors.rst
+++ b/docs/source/monitors.rst
@@ -32,8 +32,10 @@ number of items were extracted and fails if it is less than the expected thresho
                 msg='Extracted less than {} items'.format(minimum_threshold)
             )
 
-A `Monitor` instance has the following properties that can be used to help you
-implement your monitors:
+.. class:: Monitor(name=name, data=None, parent=None)
+
+A `Monitor` instance defines a monitor that includes your monitoring logic and
+has the following properties that can be used to help you implement your monitors:
 
 ``data.stats`` dict-like object containing the stats of the spider execution
 
@@ -89,8 +91,7 @@ Here is an example of how to configure a new monitor suite in your project:
 
   ``name`` suite name
 
-  ``monitors`` list of :class:`~spidermon.core.monitors.Monitor` that will be executed
-  if this suite is enabled.
+  ``monitors`` list of :class:`Monitor` that will be executed if this suite is enabled.
 
   ``monitors_finished_actions`` list of action classes that will be executed when
   all monitors finished their execution.
--- a/docs/source/monitors.rst
+++ b/docs/source/monitors.rst
@@ -32,8 +32,9 @@ number of items were extracted and fails if it is less than the expected thresho
                 msg='Extracted less than {} items'.format(minimum_threshold)
             )
 
-A `Monitor` instance defines a monitor that includes your monitoring logic and
-has the following properties that can be used to help you implement your monitors:
+A :class:`~spidermon.core.monitors.Monitor` instance defines a monitor that includes
+your monitoring logic and has the following properties that can be used to help you
+implement your monitors:
 
 ``data.stats`` dict-like object containing the stats of the spider execution
 
--- a/docs/source/monitors.rst
+++ b/docs/source/monitors.rst
@@ -91,7 +91,7 @@ Here is an example of how to configure a new monitor suite in your project:
 
   ``name`` suite name
 
-  ``monitors`` list of :class:`spidermon.core.monitors.Monitor` that will be executed if this suite is enabled.
+  ``monitors`` list of :class:`~spidermon.core.monitors.Monitor` that will be executed if this suite is enabled.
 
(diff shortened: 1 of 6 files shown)


vipulgupta2048:
There is still one more error to go, and I am positive after reading multiple issues that its an unresolved bug still in the version of Sphinx. This is the warning. 

```
Warning, treated as an error:
`/home/vipulgupta2048/spidermon/docs/source/monitors.rst:92:py:class reference target not found: spidermon.core.monitors.Monitor`
```

Gallaecio:
I think that is a valid error.

It happens if there is no reference documentation of `spidermon.core.monitors.Monitor`, and I cannot find such documentation, so the error seems valid.

Regarding how to fix it, the easiest approach your be to replace the current reference by the fully-qualified class name in monospace font:

    ``spidermon.core.monitors.Monitor``

The right solution, but maybe beyond the scope of this pull request, would be to include the reference documentation of `spidermon.core.monitors.Monitor` somewhere in the documentation. Similar to how we do it for `MonitorSuite` at https://github.com/scrapinghub/spidermon/blame/master/docs/source/monitors.rst#L85

vipulgupta2048:
@Gallaecio Thanks for the assist, I have decided to solve this permanently by adding a `Monitor` class reference in the documentation. Though, I have not added the description to arguments (namely `data` and `parent`) for I have to read the code more to understand what they are actually meant for by the developers. Let me know if this works. 

Here's the HTML build now, 
```
➜  docs git:(fix-docs) ✗ sphinx-build -nW -b html source/ {envtmpdir}/html
Running Sphinx v1.8.2
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: [] 0 added, 1 changed, 0 removed
reading sources... [100%] monitors                                                                                                                                                            
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] monitors                                                                                                                                                             
generating indices... genindex py-modindex
writing additional pages... search
copying downloadable files... [100%] ../spidermon/contrib/actions/reports/templates/reports/email/monitors/result.jinja                                                                       
copying static files... done
copying extra files... done
dumping search index in English (code: en) ... done
dumping object inventory... done
build succeeded.
```

docs/source/monitors.rst
```diff
@@ -32,8 +32,10 @@ number of items were extracted and fails if it is less than the expected thresho
                 msg='Extracted less than {} items'.format(minimum_threshold)
             )
 
-A `Monitor` instance has the following properties that can be used to help you
-implement your monitors:
+.. class:: Monitor(name=name, data=None, parent=None)
```
Gallaecio: The reference documentation should be at the end of this section, not in the middle.

Also, you should use `.. autoclass:: spidermon.core.monitors.Monitor`, followed by `:members:` in the following line, to get the documentation automatically loaded from the source code (when available).

docs/source/monitors.rst
```diff
@@ -89,8 +91,7 @@ Here is an example of how to configure a new monitor suite in your project:
 
   ``name`` suite name
 
-  ``monitors`` list of :class:`~spidermon.core.monitors.Monitor` that will be executed
-  if this suite is enabled.
+  ``monitors`` list of :class:`Monitor` that will be executed if this suite is enabled.
```
Gallaecio: After applying the change I mention above, the change in this line should be reverted, and it should work nonetheless.

docs/source/monitors.rst
vipulgupta2048: Roger that, it's my time working with class references in Sphinx. Good to know stuff!

vipulgupta2048: Where exactly do we need this section to be at, I feel this should be referenced when we actually define what are monitors.

Gallaecio: API reference usually goes at the end. First you introduce the concepts, basic usage, special scenarios, and at the end of it all goes the API reference, for people who still have questions after reading everything else about the class.

If you look at the section about the `MonitorSuite` class, it is also at the end of its section.

vipulgupta2048: Roger that, followed your advice. Let me know if this location is right, if not please let me know the exact location you would need it to be. It's just that I have a hard time figuring out where to place it best since the entire page is about monitors and placing it at the very end is weird. 

Also, should we add some documentation for the class `spidermon.core.monitors.Monitor` in the code? Seeing it empty looks a bit unusual right now. What say?

docs/source/monitors.rst
vipulgupta2048: Tick

docs/source/monitors.rst
Gallaecio: I did not meant at the end of the page, but at the end of the section. In other words, right above the Monitor Suites header.

As for documenting `spidermon.core.monitors.Monitor` in code, it definitely needs to be done eventually. It’s up to you whether you want to make it part of this changeset or leave it for a later changeset.

vipulgupta2048: @Gallaecio No worries but isn't that the same place I put the class tag at the very start :sweat_smile: Since the documentation for `spidermon.core.monitors.Monitor` is not available. I thought, keeping it above where we define the `Monitor` instance would be more appropriate. I know it doesn't look good right now but sure is better than having some explanation than no explanation. Let me know how you would like it. 
  
Screenshot_2019-10-23_10-53-26

I can surely add the docs for that in the code later, as part of another PR. WIll open an issue as well, if someone from the Hacktoberfest picks it up. Could lead to potential new contributors.

Gallaecio: I believe in your changes when I originally commented it was placed before the “A Monitor instance…” paragraph: https://github.com/scrapinghub/spidermon/commit/c113537d395421aba33c8e2ad9b7280fca032ab9

vipulgupta2048: +1 then, resolving this!

docs/source/monitors.rst
```diff
@@ -89,8 +91,7 @@ Here is an example of how to configure a new monitor suite in your project:
 
   ``name`` suite name
 
-  ``monitors`` list of :class:`~spidermon.core.monitors.Monitor` that will be executed
-  if this suite is enabled.
+  ``monitors`` list of :class:`spidermon.core.monitors.Monitor` that will be executed if this suite is enabled.
```
Gallaecio: This change should now be reverted.

docs/source/monitors.rst
```diff
@@ -32,8 +32,8 @@ number of items were extracted and fails if it is less than the expected thresho
                 msg='Extracted less than {} items'.format(minimum_threshold)
             )
 
-A `Monitor` instance has the following properties that can be used to help you
-implement your monitors:
+A `Monitor` instance defines a monitor that includes your monitoring logic and
```
Gallaecio: And this would be better now:

```suggestion
A :class:`~spidermon.core.monitors.Monitor` instance defines a monitor that includes your monitoring logic and
```

docs/source/monitors.rst
vipulgupta2048: Roger that, reverted the change. Did wanted to ask what is the `~` for? Is it something that completes the relative path of the class that we are trying to list out.

docs/source/monitors.rst
vipulgupta2048: Good catch, implemented.

vipulgupta2048:
We are good to go!

Gallaecio:
I recommend you undo your changes to `tox.ini` here, otherwise you will get conflicts if https://github.com/scrapinghub/spidermon/pull/224 is merged before this.

vipulgupta2048:
Yeap, thanks for looking out for this @Gallaecio

PR #226, opened 2019-10-31
Added tests for SendEmail action.

felipefin:
Added unit tests for SendEmail action.


tests/contrib/actions/email/test_email_action.py
```diff
@@ -0,0 +1,195 @@
...
+    "sender,recipient,fake,subject,exception_message",
+    [
+        (
+            None,
+            None,
+            None,
+            None,
+            "You must provide at least one recipient for the message.",
+        ),
+        (SENDER, RECIPIENT, None, None, "You must provide a subject for the message."),
```
Gallaecio: What about having a `SENDER` but no `RECIPIENT`? I don’t see that in either `test_init_with_invalid_params` or `test_init_with_valid_params`.

tests/contrib/actions/email/test_email_action.py
```diff
@@ -0,0 +1,195 @@
...
+    email.fake = False
+    email.run_action()
+    email.get_message.assert_called_once_with()
+    logger.info.assert_not_called()
+    email.send_message.assert_called_once_with(email.get_message())
+
+
+def test_get_subject_with_subject_present(email):
+    email.get_subject()
+    email.render_text_template.assert_called_once_with(SUBJECT)
```
Gallaecio: I think we should use the object API whenever possible instead of a mock, and leave mocks only for scenarios that cannot be tested otherwise.

For a test like this, I think it would make more sense to create an Email object, and simply check that the subject passed to the ``__init__`` method matches the one that `email.get_subject()` returns.

tox.ini
```diff
@@ -3,7 +3,8 @@ envlist = py27,py35,py36,py37
 skip_missing_interpreters = True
 
 [testenv]
-extras =
+deps = -r {toxinidir}/requirements.txt
```
Gallaecio: Doesn’t this work just the same?

```suggestion
deps = -r requirements.txt
```

tests/contrib/actions/email/test_email_action.py
FelipeFin: Thanks for the suggestion. It seems that I missed that test.

tests/contrib/actions/email/test_email_action.py
FelipeFin: My opinion is that we should isolate each part of the function to test all the cases.
For instance, in the test above, it seems that it does not matter the outcome of render_text_template (since the return of this method should be tested elsewhere), just that it is called with the correct argument.
However I'm open for discussion :)

rennerocha: I agree with @Gallaecio . Isolate things is important, but in this case it seems you are mocking internal behavior of the class that can be changed anytime without affecting the purpose of the test. Usually I use mocks only when I am dealing with external objects/systems.

PR #229, opened 2019-11-14
Added tests for JobTagsAction

felipefin:
Added unit tests for **JobTagsAction**, **AddJobTags** and **RemoveJobTags** actions.


tests/contrib/actions/jobs/test_tags.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.mark.parametrize(
+    "tags, expected_tags", [(None, []), ([TAG], [TAG]), (TAG, [TAG])],
+)
+def test_job_tags_action_init(tags, expected_tags):
+    job_tags = JobTagsAction(tags)
+    assert job_tags.tags == expected_tags
+
+
+class JobTagsActionTests:
```
Gallaecio: Why the class here?

FelipeFin: The intention is to keep the code more organized, given the fact the all classes have methods with same name.
Do you suggest another approach?

PR #230, opened 2019-11-19
chore: Added unit tests for reports actions

felipefin:
Added unit tests for reports actions

### DIFF
--- a/tests/contrib/actions/reports/test_files.py
+++ b/tests/contrib/actions/reports/test_files.py
@@ -1,9 +1,4 @@
-try:
-    from unittest.mock import MagicMock, patch, mock_open
-except ImportError:
-    from mock import MagicMock, patch, mock_open
 import pytest
-from unittest import TestCase
 
 from spidermon.contrib.actions.reports.files import CreateFileReport
 from spidermon.exceptions import NotConfigured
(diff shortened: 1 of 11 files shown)


tests/contrib/actions/reports/test_files.py
```diff
@@ -0,0 +1,43 @@
+try:
+    from unittest.mock import MagicMock, patch, mock_open
+except ImportError:
+    from mock import MagicMock, patch, mock_open
```
rennerocha: You can use `pytest-mock` library to avoid this kind of imports.

tests/contrib/actions/reports/test_files.py
```diff
@@ -0,0 +1,43 @@
...
+from unittest import TestCase
+
+from spidermon.contrib.actions.reports.files import CreateFileReport
+from spidermon.exceptions import NotConfigured
+
+TEMPLATE = "template"
+FILENAME = "filename"
+
+
+class CreateFileReportTests(TestCase):
```
rennerocha: `pytest` doesn't require a `TestCase` class for the tests. All tests in Spidermon are single function.

tests/contrib/actions/reports/test_reports.py
```diff
@@ -0,0 +1,40 @@
+try:
+    from unittest.mock import MagicMock, patch, call, mock_open
```
rennerocha: You can use `pytest-mock` library (and the `mocker` fixture) as you did in `test_files.py` to keep consistency.

tests/contrib/actions/reports/test_reports.py
```diff
@@ -0,0 +1,40 @@
...
+from unittest import TestCase
+
+from spidermon.contrib.actions.reports import CreateReport
+from spidermon.exceptions import NotConfigured
+
+TEMPLATE = "template"
+FILENAME = "filename"
+
+
+class CreateReportTests(TestCase):
```
rennerocha: Sometime it could be useful, but to keep consistency over the other tests, usually we don't have a `TestCase` class for our tests.

tests/contrib/actions/reports/test_s3.py
```diff
@@ -0,0 +1,201 @@
+try:
+    from unittest.mock import MagicMock, patch, mock_open
```
rennerocha: You can use `pytest-mock` library (and the `mocker` fixture) as you did in test_files.py to keep consistency.

PR #133, opened 2019-03-23
docs: Mixins documentation

gcmurillo:
Adding docs for mixin objects in `spidermincontrib/monitors/mixins`

PR #232, opened 2019-11-26
docs: Documenting built-in mixins

rennerocha:
Continue work started on https://github.com/scrapinghub/spidermon/pull/133


mirceachira:
Hey @rennerocha I've looking a bit into #133 and this and was wondering if the issue could be up for grabs?

I think documenting SpiderMonitorMixin should include some docs for `spidermon.contrib.stats.counters.DictPercentCounter` which is mixed with ResponseInfo from mixin class.

 I'd be happy to contribute some docs to this.

PR #234, opened 2019-11-28
feature: Allowed to define SPIDERMON_SLACK_RECIPIENTS setting as a comma-separated string with the desired recipients.

### LINKED ISSUE #233: It's not possible to specify SPIDERMON_SLACK_RECIPIENTS via Scrapy Cloud
We should replace `settings.get` by `settings.getlist`. I'm working on a pull request.

victor-torres:
This PR should fix #233

### DIFF
--- a/tests/contrib/actions/slack/test_slack_message_manager.py
+++ b/tests/contrib/actions/slack/test_slack_message_manager.py
@@ -1,5 +1,17 @@
 import pytest
-from spidermon.contrib.actions.slack import SlackMessageManager
+
+from scrapy.crawler import CrawlerRunner
+from scrapy.spiders import Spider
+
+from spidermon.contrib.actions.slack import (
+    SlackMessageManager,
+    SendSlackMessage,
+)
+
+
+def get_crawler(settings: dict = None):
+    runner = CrawlerRunner(settings)
+    return runner.create_crawler(Spider)
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


victor-torres:
cc @ejulio

tests/contrib/actions/slack/test_slack_message_manager.py
```diff
@@ -1,5 +1,17 @@
...
+from scrapy.crawler import CrawlerRunner
+from scrapy.spiders import Spider
+
+from spidermon.contrib.actions.slack import (
+    SlackMessageManager,
+    SendSlackMessage,
+)
+
+
+def get_crawler(settings=None):
```
rennerocha: You can use `scrapy.utils.test.get_crawler` function available in Scrapy (which is pretty much the same you did), so we can avoid problems if Scrapy changes the way a test instance of crawler is done.

tests/contrib/actions/slack/test_slack_message_manager.py
```diff
@@ -39,3 +51,12 @@ def test_api_call_with_error_should_log_error_msg(mocker, logger_error):
 
     assert logger_error.call_count == 1
     assert error_msg in logger_error.call_args_list[0][0]
+
+
+def test_load_recipients_list_from_crawler_settings():
+    settings = {
+        "SPIDERMON_SLACK_RECIPIENTS": "foo,bar",
```
rennerocha: I believe it is worthy to include a test case for the already existing behaviour (`["foo", "bar"]`) , so we can be sure we didn't break anything. You can use `@pytest.mark.parametrize` to avoid duplicating the test code.

victor-torres:
Thank you for your suggestions, @rennerocha. I've just pushed a couple of commits that should have addressed them. Let me know if I can help with anything else.

PR #236, opened 2019-12-27
chore: Configure test coverage reporting

rennerocha:


vipulgupta2048:
Looking great! It also tracks Travis Builds alongside it.  
Does ScrapingHub have a team `Codecov yaml` or should we add one for further customization?

Issue #239, opened 2020-01-24
SPIDERMON_ENABLED not disabling spidermon related item pipelines


rennerocha:
Even if we set `SPIDERMON_ENABLED` to `False`, item validation pipeline still works, so spidermon is not completely disabled by this setting.

PR #240, opened 2020-01-27
feature: Disable item validation pipeline when Spidermon is disabled

rennerocha:
This may fix https://github.com/scrapinghub/spidermon/issues/239


spidermon/contrib/scrapy/pipelines.py
```diff
@@ -41,6 +46,10 @@ def __init__(
 
     @classmethod
     def from_crawler(cls, crawler):
+        spidermon_enabled = crawler.settings.getbool("SPIDERMON_ENABLED")
+        if not spidermon_enabled:
```
muzaffaryousaf: Would it disable for those projects where we have not set SPIDERMON_ENABLED flag explicitly ?

rennerocha: Yes, it will. I am considering that item validation pipeline is a spidermon related pipeline, so if we don't explicitly enable spidermon, everything spidermon related should not be executed.

We had this problem in a project where spidermon was disabled in production but the pipeline was there with SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS set to True. This leads to production items with `_validation` field populated which is not desired.

spidermon/contrib/scrapy/pipelines.py
```diff
@@ -41,6 +46,10 @@ def __init__(
...
+            return PassThroughPipeline()
```
ejulio: Shouldn't it `raise NotConfigured`?

spidermon/contrib/scrapy/pipelines.py
ejulio: The question I get here is, does it make sense to have only monitoring or validation enabled?
Say, is there a case where I just monitoring, but not validation?
Or, is there a case I just want validation, but no monitoring?
If so, I'd go for two setting instead of one.
Otherwise, if we consider spidermon as a whole, then this is the preferred solution

rennerocha: Yes, it does make sense to use just the validation feature of Spidermon or just the monitoring part. However in this case I am considering Spidermon as a whole, so we should be able to enable/disable everything Spidermon related using just one setting to avoid problems like the one I mentioned.

PR #241, opened 2020-01-30
bug: Fixed Telegram action error logging

anapaulagomes:
If the request sent to Telegram throw an error the following dict is returned: 

```
{'ok': False, 'error_code': 400, 'description': 'Bad Request: chat not found'}
```

Currently, Spidermon is calling the key `text` instead of `description` in order to add details to the logs, which is causing the following error:

```
ERROR: SendTelegramMessageSpiderFinished
  File "/home/ana/workspace/xxx/venv/lib/python3.8/site-packages/spidermon/core/actions.py", line 37, in run
    self.run_action()
  File "/home/ana/workspace/xxx/venv/lib/python3.8/site-packages/spidermon/contrib/actions/telegram/__init__.py", line 94, in run_action
  File "/home/ana/workspace/xxx/venv/lib/python3.8/site-packages/spidermon/contrib/actions/telegram/__init__.py", line 48, in send_message
  File "/home/ana/workspace/xxx/venv/lib/python3.8/site-packages/spidermon/contrib/actions/telegram/__init__.py", line 28, in send_message
    logger.error("Failed to send message. Telegram api error: %s", r.text)
AttributeError: 'dict' object has no attribute 'text'
```

This PR fixes it.

**Version:** Spidermon 1.12.0

### DIFF
--- a/spidermon/contrib/actions/telegram/__init__.py
+++ b/spidermon/contrib/actions/telegram/__init__.py
@@ -25,7 +25,7 @@ class SimplyTelegramClient:
         )
         r = requests.get(api_url).json()
         if r.get("ok") is False:
-            logger.error("Failed to send message. Telegram api error: %s", r.text)
+            logger.error("Failed to send message. Telegram api error: %s", r.description)
 
 
 class TelegramMessageManager:
--- a/spidermon/contrib/actions/telegram/__init__.py
+++ b/spidermon/contrib/actions/telegram/__init__.py
@@ -25,7 +25,9 @@ class SimplyTelegramClient:
         )
         r = requests.get(api_url).json()
         if r.get("ok") is False:
-            logger.error("Failed to send message. Telegram api error: %s", r.description)
+            logger.error(
+                "Failed to send message. Telegram api error: %s", r.description
+            )
 
 
 class TelegramMessageManager:
--- a/spidermon/contrib/actions/telegram/__init__.py
+++ b/spidermon/contrib/actions/telegram/__init__.py
@@ -26,7 +26,7 @@ class SimplyTelegramClient:
         r = requests.get(api_url).json()
         if r.get("ok") is False:
             logger.error(
-                "Failed to send message. Telegram api error: %s", r.description
+                "Failed to send message. Telegram api error: %s", json.dumps(r)
             )
 
 
--- a/tests/contrib/actions/telegram/test_telegram_action.py
+++ b/tests/contrib/actions/telegram/test_telegram_action.py
@@ -1,3 +1,4 @@
+import json
 import pytest
 
 from scrapy import Spider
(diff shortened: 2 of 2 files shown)


Sammeeey:
I am having the same issue.
I installed scrapy with `pip install scrapy`.
Then I have just been following the [Getting started](https://spidermon.readthedocs.io/en/latest/getting-started.html) Tutorial.
When changing the `minimum_threshold` to 200 I get the error message above.

So I changed the fix in _init_ manually but it does not work for me.

spidermon/contrib/actions/telegram/__init__.py
```diff
@@ -25,7 +25,9 @@ def send_message(self, message, recipient):
         )
         r = requests.get(api_url).json()
         if r.get("ok") is False:
-            logger.error("Failed to send message. Telegram api error: %s", r.text)
+            logger.error(
+                "Failed to send message. Telegram api error: %s", r.description
```
Gallaecio: Could you add a test for this change?

I believe this will simply cause the error message to change to `AttributeError: 'dict' object has no attribute 'description'`. The fix is probably to change `r.description` to `r['description']`.

Beyond that, I’m not aware of whether `text` or `description` are the right choice. Looking at the original code, I suspect the intention was to log the whole response (and the error was not to take into account that `.json()` was used). In that case, I think the best option would be to change `r.text` to `json.dumps(r)`.

Sammeeey: **Fix for action log**:

@Gallaecio you are right, the following would be the error message for [the recommended change](https://github.com/scrapinghub/spidermon/pull/241/commits/a79c82b09fecaf31a343ca508ba9167342037eaa):
```
ERROR: SendTelegramMessageSpiderFinished
  File "c:\users\samuel\anaconda3\lib\site-packages\spidermon\core\actions.py", line 37, in run
    self.run_action()
  File "c:\users\xxxx\anaconda3\lib\site-packages\spidermon\contrib\actions\telegram\__init__.py", line 95, in run_action
    self.manager.send_message(to=self.recipients, text=self.get_message())
  File "c:\users\xxxx\anaconda3\lib\site-packages\spidermon\contrib\actions\telegram\__init__.py", line 49, in send_message
    self._client.send_message(text, recipient)
  File "c:\users\xxxx\anaconda3\lib\site-packages\spidermon\contrib\actions\telegram\__init__.py", line 29, in send_message
    "Failed to send message. Telegram api error: %s", r.description
AttributeError: 'dict' object has no attribute 'description'
```

Your following [suggestion](https://github.com/scrapinghub/spidermon/pull/241#discussion_r386129261) seems to be the solution for the problem above.
>The fix is probably to change r.description to r['description'].
This works for me.

Thank you very much!

anapaulagomes: Hummm shouldn't the build failed in this case? I'll add a test and make the change anyway. Thanks for your input.

PR #242, opened 2020-02-06
feature: Item validation built in monitor

rennerocha:
It is common when using item validation pipelines to create a monitor that verifies if there are validation errors in the job results (`spidermon/validation/fields/errors` stat). This PR includes a basic monitor to be included in the basic monitor suite available in spidermon.

### DIFF
--- a/spidermon/contrib/scrapy/monitors.py
+++ b/spidermon/contrib/scrapy/monitors.py
@@ -1,6 +1,7 @@
-from spidermon import monitors, MonitorSuite, Monitor
+from spidermon import Monitor, MonitorSuite, monitors
 from spidermon.exceptions import NotConfigured
 from spidermon.utils.settings import getdictorlist
+
 from ..monitors.mixins.spider import SpiderMonitorMixin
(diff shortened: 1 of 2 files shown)


spidermon/contrib/scrapy/monitors.py
```diff
@@ -153,10 +155,28 @@ def test_check_unwanted_http_codes(self):
...
+
+    You can configure the maximum number of item validation errors using
+    ``SPIDERMON_MAX_ITEM_VALIDATION_ERRORS``. The default is ``0``."""
+
+    @monitors.name("Should not have more item validation errors than configured.")
+    def test_verify_item_validation_error(self):
+        errors_threshold = self.crawler.settings.getint(
+            SPIDERMON_MAX_ITEM_VALIDATION_ERRORS, 0
+        )
+        item_validation_errors = self.stats.get("spidermon/validation/fields/errors", 0)
```
victor-torres: Should we use getint or cast the result to an integer?

PR #248, opened 2020-05-06
bug: AttributeError when using ValidationMonitorMixin

### LINKED ISSUE #246: bug: AttributeError when using ValidationMonitorMixin
When executing a monitor using `ValidationMonitorMixin`, a `AttributeError` is raised because a needed attribute is not set:
`AttributeError: 'CoverageValidationMonitor' object has no attribute 'correct_field_list_handling'`

This answers this comment:
> @ejulio I've implemented that and wrote tests for all methods for both old and new behavior.
> 
> My only concern about setting the flag in `__init__`: as this is a mixin I'm not sure how should this be used in user code. In the test I just set the variable instead of calling mixin's `__init__`.
> 
> _Originally posted by @wRAR in https://github.com/scrapinghub/spidermon/pull/164#issuecomment-508067602_

rennerocha:
This PR fixes #246 .

### DIFF
--- a/spidermon/contrib/monitors/mixins/validation.py
+++ b/spidermon/contrib/monitors/mixins/validation.py
@@ -86,9 +86,8 @@ class ValidationInfo(object):
 
 
 class ValidationMonitorMixin(StatsMonitorMixin):
-    def __init__(self, correct_field_list_handling=False):
-        super(ValidationMonitorMixin, self).__init__()
-        self.correct_field_list_handling = correct_field_list_handling
+
+    correct_field_list_handling = False
 
     @property
     def validation(self):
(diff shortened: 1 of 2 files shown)


Gallaecio:
https://github.com/scrapinghub/spidermon/issues/246 sounds to me like a bad usage in the code using the mixing (defining `__init__` without calling `super().__init__`), so I’m not sure it is a bug. But I guess the change results in better usability. So +0 from me.

rennerocha:
> #246 sounds to me like a bad usage in the code using the mixin (defining `__init__` without calling `super().__init__`), so I’m not sure it is a bug. But I guess the change results in better usability. So +0 from me.

Mixin `__init__` is never called so that is way we have this bug. I agree that this is not the right usage for this attribute anyway. Maybe in a next release we can get rid of it. For now, let's make the mixin usable.

Gallaecio:
I’m not walking about using `super` in the mixin, but in the classes using the mixin.

Whenever you inherit from a class, be it a mixin or something else, you are still inheriting from a class, and you must use `super` in your `__init__`. Otherwise, you are preventing the `__init__` of parent classes to be executed.

This specific mixin can be modified not to use `__init__`, but that will not always be the case.

PR #252, opened 2020-05-21
chore: Removed Python 2.7 support

rennerocha:
…ed code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ python:
   - 3.5
   - 3.6
   - 3.7
-  - 3.8
 install:
   - pip install tox-travis codecov
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - 3.5
   - 3.6
   - 3.7
+  - 3.8
 install:
   - pip install tox-travis codecov
 script:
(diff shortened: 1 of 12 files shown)


Gallaecio:
Since the documentation says 3.5+, should this patch also include official support for 3.8?

rennerocha:
Build is passing:
https://travis-ci.org/github/scrapinghub/spidermon/builds/690104886

wRAR:
`six` is still listed at least in setup.py and requirements.txt. Or is it still used by some code?

rennerocha:
> `six` is still listed at least in setup.py and requirements.txt. Or is it still used by some code?

`six` is still being used in several places. I only removed parts of code that are explicitly related to Python 2 (places where we find `if six.PY2` for example). I was a bit conservative to avoid removing more code than needed.

.travis.yml
```diff
@@ -5,7 +5,6 @@ git:
 language: python
 dist: xenial
 python:
-  - 2.7
   - 3.5
   - 3.6
   - 3.7
```
Gallaecio, line 10:
```suggestion
  - 3.7
  - 3.8
```

rennerocha:
Build is passing:
https://travis-ci.org/github/scrapinghub/spidermon

I removed Python 3.8 as it was breaking the build and it will need a better investigation how to fix it.
This PR will only contain changes necessary to drop Python 2.7 support. A new issue will be created to add Python 3.8 support.

Issue #251, opened 2020-05-20
Update slackclient version


rennerocha:
`SendSlackMessage` action don't work with the 2.x `slackclient` versions. To avoid breaking spidermon, we are locked with 1.x version.

There is a guide for migrating the library as it was completely rewritten (so it is backward incompatible):
https://github.com/slackapi/python-slackclient/wiki/Migrating-to-2.x

This was related to this issue:
https://github.com/scrapinghub/scrapinghub-stack-scrapy/issues/141

andersonberg:
I will start working on this issue

rennerocha:
Thanks @andersonberg ! Please let me know if you need some assistance.

PR #260, opened 2020-07-13
fix: use users.info instead of users.list

karenyavine:
The [users.list](https://api.slack.com/methods/users.list) Slack api endpoint returns information about all users in a given workspace, which could possibly lead to data leakage.

Since spidermon only needs information about it's own bot user, this change retrieves the user id from an [auth.test](https://api.slack.com/methods/auth.test) call, and then call [users.info](https://api.slack.com/methods/users.info), reducing the data retrieved from Slack.

Unfortunately the Slack API scopes are not granular enough to also restrict the token to users.info and not users.list (both are covered by users:read scope), but this change reduces the risk of data leakage in logs or memory dumps.

Additionally, this moves the check on if the request got errors to a separate function

### DIFF
--- a/spidermon/contrib/actions/slack/__init__.py
+++ b/spidermon/contrib/actions/slack/__init__.py
@@ -84,21 +84,23 @@ class SlackMessageManager:
         user = self.users.get(name, None)
         return user["id"] if user else None
 
-    def _get_users_info(self):
-        return dict(
-            [
-                (member["name"].lower(), member)
-                for member in self._api_call("users.list")["members"]
-            ]
-        )
-
-    def _api_call(self, method, **kwargs):
-        response = self._client.api_call(method, **kwargs)
-
+    def _req_had_errors(self, response):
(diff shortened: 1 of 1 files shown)


rennerocha:
Thanks @karenyavine for you PR. There is an [open issue](https://github.com/scrapinghub/spidermon/issues/251) to update `slackclient` version of this action, and I am not sure if we will need to modify this part of code.

@andersonberg , as you are working on this issue now, could you please verify if we will have some change here or if this PR can be applied?

rennerocha:
@karenyavine ,

Thanks again for you contribution. It is really important to avoid obtaining more data than necessary and I would love if this is possible thorugh Slack API. However, with this change, it won't be possible to send Direct Messages to a user [as commented here](https://github.com/scrapinghub/spidermon/pull/265#discussion_r495049108).

After upgrading to a newer `slackclient` version we may search a better way to handle this scenario.

Issue #253, opened 2020-05-22
feature: Built-in monitor for field coverage


rennerocha:
In certain spiders we can accept that not all items returned contains a certain field, but we expect that at least certain amount of items returned contains this field. We can use `ValidationMonitorMixin` to create monitors that do this kind of validation. However this requires us to write code and updating the thresholds requires changes in the source code.

My idea is to create a new built-in monitor, configured by a new setting that defines all item coverage rules in a more straightforward way. When an user defines this setting and enable the monitor, we will be notified when the rules weren't satisfied.

Example of proposed setting:
We have `myproject.MyItem`, the spider returns 100 of this item and we have the following configuration:

```
SPIDERMON_FIELD_COVERAGE = {
    "myproject.MyItem": {
        "field_name_1": {
            # Minimum percent of items that needs to have this field
            "min_percent": 10,
        },
        "field_name_2": {
            # Minimum number of items that needs to have this field
            "min_items": 5,
        },
    }
}
```

The new monitor will fail if:
- We have more than 10% of the items missing `field_name_1`; or
- We have more than 5 items missing `field_name_2`.

Looking for comments over this proposal to improve it.

simoess:
@rennerocha I think the proposed format will be a good format to standardize the way that the field coverage monitors are defined. A built-in monitor would be great. On most project's that I've worked on, the settings are usually defined this way:

SPIDER_NAME: {
    "field_name_1": min_percent,
    "field_name_2": min_percent,
    "composed_field_name": { # only one project that I've worked on had the option to define field coverage for composed fields
          "sub_field_name_1": min_percent
    }
}

It would be nice to be able to define monitors for  item coverage and field coverage in one place.

Gallaecio:
For composed fields, in a project where I used to work we would use the “field path”, e.g. “composed_field_name/sub_field_name_1”.

Issues to consider:
- Fields containing `/` in their name would be a problem.
- Should relative values (percentages) be relative to the parent field or to the item count?
- List item count checks. If `field1` is a list, and you want to check totals of `len(field1)`.

rennerocha:
`ValidationMonitorMixin` implementation requires that to be able to do this kind of validation, we need to set a field as required, even if it is not, or else we will not have the stats of the coverage. This new monitor should remove this need, so we can have fields that are not required (in the JSON schema for example), but we need some coverage.

PTaylour:
[...]

PR #262, opened 2020-07-30
feature: Add field coverage statistics

rennerocha:
After this PR is merged, it will be easier to implement #253 as a new built-in monitor for field coverage.

When `SPIDERMON_ADD_FIELD_COVERAGE` is enabled, it will include a new set of stats containing information about field coverage for different item types, making it possible to use these stats in custom monitors.


andersonberg:
Good work on the tests

PR #263, opened 2020-07-31
feature: Built-in monitor for field coverage

rennerocha:
This PR resolves https://github.com/scrapinghub/spidermon/issues/253

This PR needs to be merged together or else this PR won't work:
https://github.com/scrapinghub/spidermon/pull/262

### DIFF
--- a/docs/source/monitors.rst
+++ b/docs/source/monitors.rst
@@ -1,5 +1,3 @@
-.. _monitors:
-
 ====================
 Monitoring your jobs
 ====================
@@ -138,7 +136,7 @@ The Basic Monitors
 Spidermon has some batteries included :)
 
 .. automodule:: spidermon.contrib.scrapy.monitors
-    :members: FinishReasonMonitor, ItemCountMonitor, ErrorCountMonitor, UnwantedHTTPCodesMonitor
+    :members: ErrorCountMonitor, FieldCoverageMonitor, FinishReasonMonitor, ItemCountMonitor, UnwantedHTTPCodesMonitor
 
 Is there a **Basic Scrapy Suite** ready to use?
 ------------------------------------------------
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -123,7 +123,7 @@ SPIDERMON_ADD_FIELD_COVERAGE
 Default: ``False``
 
 When enabled, Spidermon will add statistics about the number of items scraped and coverage for each existing
-field following the format:
+field following this format:
 
 ``'spidermon_item_scraped_count/<item_type>/<field_name>': <item_count>``
 ``'spidermon_field_coverage/<item_type>/<field_name>': <coverage>``
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -167,7 +167,7 @@ SPIDERMON_FIELD_COVERAGE_SKIP_NONE
 ----------------------------------
 Default: ``False``
 
-When enabled, fields returned that contains `None` as value will not be counted as an existing value.
+When enabled, returned fields that have ``None`` as value will not be counted as fields with a value.
 
 Considering your spider returns the following items:
 
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -116,3 +116,12 @@ SPIDERMON_ENGINE_STOP_EXPRESSION_MONITORS
 Default: ``[]``
 
 List of dictionaries describing :ref:`expression monitors<topics-expression-monitors>` to run when the engine is stopped.
+
+
+.. SPIDERMON_ADD_FIELD_COVERAGE:
+
+SPIDERMON_ADD_FIELD_COVERAGE
+----------------------------
+Default: ``False``
+
+When enabled, Spidermon will add field coverage information in Spider stats.
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -124,4 +124,44 @@ SPIDERMON_ADD_FIELD_COVERAGE
 ----------------------------
 Default: ``False``
 
-When enabled, Spidermon will add field coverage information in Spider stats.
+When enabled, Spidermon will add statistics about the number of items scraped and coverage for each existing
+field following the format:
+
+``'spidermon_item_scraped_count/<item_type>/<field_name>': <coverage>``
+``'spidermon_field_coverage/<item_type>/<field_name>': <item_count>``
+
+.. note::
+
+   Nested fields are also supported. For example, if your spider returns these items:
+
+   .. code-block:: python
+
+      [
+        {
+          "field_1": {
+            "nested_field_1_1": "value",
+            "nested_field_1_2": "value",
+          },
+        },
+        {
+          "field_1": {
+            "nested_field_1_1": "value",
+          },
+          "field_2": "value"
+        },
+      ]
+
+   Statistics will be like the following:
+
+   .. code-block:: python
+
+      'spidermon_item_scraped_count/dict': 2,
+      'spidermon_item_scraped_count/dict/field_1': 2,
+      'spidermon_item_scraped_count/dict/field_1/nested_field_1_1': 2,
+      'spidermon_item_scraped_count/dict/field_1/nested_field_1_2': 1,
+      'spidermon_item_scraped_count/dict/field_2': 1,
+      'spidermon_field_coverage/dict/field_1': 1,
+      'spidermon_field_coverage/dict/field_1/nested_field_1_1': 1,
+      'spidermon_field_coverage/dict/field_1/nested_field_1_2': 0.5,
+      'spidermon_item_scraped_count/dict/field_2': 0.5,
+
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -165,3 +165,30 @@ field following the format:
       'spidermon_field_coverage/dict/field_1/nested_field_1_2': 0.5,
       'spidermon_item_scraped_count/dict/field_2': 0.5,
 
+.. note::
+
+   A field is counted as scraped and covered if it is returned by the spider, no matter
+   its content. So for the following set of items:
+
+   .. code-block:: python
+
+      [
+        {
+          "field_1": None,
+          "field_2": "value",
+        },
+        {
+          "field_1": "value",
+          "field_2": "value",
+        },
+      ]
+
+    Statistics will be like the following:
+
+   .. code-block:: python
+
+      'spidermon_item_scraped_count/dict': 2,
+      'spidermon_item_scraped_count/dict/field_1': 2,
+      'spidermon_item_scraped_count/dict/field_2': 2,
+      'spidermon_field_coverage/dict/field_1': 1,
+      'spidermon_item_scraped_count/dict/field_2': 1,
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -118,8 +118,6 @@ Default: ``[]``
 List of dictionaries describing :ref:`expression monitors<topics-expression-monitors>` to run when the engine is stopped.
 
 
-.. SPIDERMON_ADD_FIELD_COVERAGE:
-
 SPIDERMON_ADD_FIELD_COVERAGE
 ----------------------------
 Default: ``False``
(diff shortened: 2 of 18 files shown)


docs/source/monitors.rst
```diff
@@ -1,5 +1,3 @@
-.. _monitors:
```
Gallaecio, line 1: Why remove this?

It’s not wrong to remove it if it is not used internally and it is _probably_ not used externally yes (through InterSphinx). However, it’s good to have references so that future projects may use InterSphinx to link to this section without having to rely on the documentation page source file path.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
         self.assertTrue(item_validation_errors <= errors_threshold, msg=msg)
 
 
+@monitors.name("Field Coverage Monitor")
+class FieldCoverageMonitor(BaseScrapyMonitor):
+    """Validate if field coverage rules are met.
+
+    To use this monitor you need to enable ``SPIDERMON_ADD_FIELD_COVERAGE``
+    setting that will add information about field coverage on your spider
+    statistics.
```
Gallaecio, line 183:
```suggestion
    To use this monitor you need to enable the ``SPIDERMON_ADD_FIELD_COVERAGE``
    setting, which will add information about field coverage to your spider
    statistics.
```

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+
+    To define your field coverage rules create a dictionary containing the
+    expected coverage for each field you want to monitor.
+
+    As example, if the items you are returning from your spider are a Python dictionaries
+    with the following format:
```
Gallaecio, line 189:
```suggestion
    As an example, if the items you are returning from your spider are Python dictionaries
    with the following format:
```

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+        # project/settings.py
+        SPIDERMON_FIELD_COVERAGE_RULES = {
+            "dict/field_1": 0.4,  # Expected 40% coverage for field_1
+            "dict/field_2": 1.0,  # Expected 100% coverage for field_2
+            "dict/field_3": 1.0,  # Expected 100% coverage for parent field_3
+            "dict/field_3/field_3_1": 0.5,  # Expected 50% coverage for nested field_3_1
+        }
+
+    You are not obligated to set rules for every field, just the ones thar you are interested.
+    Also, you can monitor nested fields if available in your returned items.
```
Gallaecio, line 215:
```suggestion
    You are not obligated to set rules for every field, just for the ones in which you are interested.
    Also, you can monitor nested fields if available in your returned items.
```

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+               "MyCustomItem/field_2": 1.0,
+           }"""
+
+    def run(self, result):
+        add_field_coverage_set = self.crawler.settings.getbool(
+            "SPIDERMON_ADD_FIELD_COVERAGE", False
+        )
+        if not add_field_coverage_set:
+            raise NotConfigured(
+                "You should enable field coverage stats (SPIDERMON_ADD_FIELD_COVERAGE=True) on your project settings"
```
Gallaecio, line 244: I’m not completely sure about the context where this appears to the user, as I’m not familiar with how Spidermon monitors are enabled and configured, so my suggestion below may be completely wrong.

But, assuming that this message appears when users have actively added the monitor somewhere, but might either have forgotten to set this setting or may have set this setting to `False` on purpose (e.g. to temporarily disable the monitor without removing it from where they added it), I would avoid this “should” phrasing.

```suggestion
                "To enable field coverage stats, set SPIDERMON_ADD_FIELD_COVERAGE=True in your project settings"
```

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+            )
+        return super(FieldCoverageMonitor, self).run(result)
+
+    def test_check_if_field_coverage_rules_are_met(self):
+        failures = []
+        field_coverage_rules = self.crawler.settings.getdict(
+            "SPIDERMON_FIELD_COVERAGE_RULES", []
```
Gallaecio, line 251:
```suggestion
            "SPIDERMON_FIELD_COVERAGE_RULES"
```

The current code works (it returns a dictionary, not a list, if the setting is undefined), but this works as well and should be less confusing.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+        )
+        for field, expected_coverage in field_coverage_rules.items():
+            real_coverage = self.data.stats.get(
```
Gallaecio, line 254: I think “real” is a bit of a false friend, “actual” is usually used in opposition to “expected”.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+                "spidermon_field_coverage/{}".format(field), 0
+            )
+            if real_coverage < expected_coverage:
+                failures.append(
+                    "{} (expected {} / real {})".format(
```
Gallaecio, line 259: :lipstick: Maybe “expected {}, got {}”?

Also, I wonder whether or not it would be better to use a percentage here. I’m unsure myself.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+                        field, expected_coverage, real_coverage
+                    )
+                )
+
+        msg = "\nField coverage rules were not met on the following fields:\n{}".format(
```
Gallaecio, line 264: I think this is a scenario where “in” should be used instead of “on”, but since I always get those 2 confused, I suggest switching the passive sentence to an active one: 

```suggestion
        msg = "\nThe following items did not meet field coverage rules:\n{}".format(
```

docs/source/monitors.rst
rennerocha: When building the docs, sphinx complained that this is a duplicated reference. We have `Monitors` as title some lines below, so both shares the same reference name. So removing this will not affect existing internal or external references.

spidermon/contrib/scrapy/monitors.py
rennerocha: As we need to define the values between in the 0-1 range, I believe it would be clearer to keep that way as well. I don't think it will be confusing.

rennerocha:
> Maybe this is not the right pull request for it, but I think we need to address in the documentation what is Spidermon’s behavior regarding list items nested field coverage. e.g.
> 
> ```
> {
>     "field1": [
>         {
>             "subfield1": "value".
>         },
>         {},
>     ]
> }
> ```
> 
> How is the coverage of "field1/subfield1" reported?

I think the best location to add this information is inside `settings.rst` describing `SPIDERMON_ADD_FIELD_COVERAGE` setting. I will add there.

docs/source/settings.rst
```diff
@@ -124,4 +124,44 @@ SPIDERMON_ADD_FIELD_COVERAGE
 ----------------------------
 Default: ``False``
 
-When enabled, Spidermon will add field coverage information in Spider stats.
+When enabled, Spidermon will add statistics about the number of items scraped and coverage for each existing
+field following the format:
+
+``'spidermon_item_scraped_count/<item_type>/<field_name>': <coverage>``
+``'spidermon_field_coverage/<item_type>/<field_name>': <item_count>``
```
Gallaecio, line 131:
```suggestion
``'spidermon_item_scraped_count/<item_type>/<field_name>': <item_count>``
``'spidermon_field_coverage/<item_type>/<field_name>': <coverage>``
```

docs/source/settings.rst
```diff
@@ -124,4 +124,44 @@ SPIDERMON_ADD_FIELD_COVERAGE
...
+
+      'spidermon_item_scraped_count/dict': 2,
+      'spidermon_item_scraped_count/dict/field_1': 2,
+      'spidermon_item_scraped_count/dict/field_1/nested_field_1_1': 2,
+      'spidermon_item_scraped_count/dict/field_1/nested_field_1_2': 1,
+      'spidermon_item_scraped_count/dict/field_2': 1,
+      'spidermon_field_coverage/dict/field_1': 1,
+      'spidermon_field_coverage/dict/field_1/nested_field_1_1': 1,
+      'spidermon_field_coverage/dict/field_1/nested_field_1_2': 0.5,
+      'spidermon_item_scraped_count/dict/field_2': 0.5,
```
Gallaecio, line 166: This covers “regular” nested fields. What about [nested fields within lists](https://github.com/scrapinghub/spidermon/pull/263#pullrequestreview-459565813)? (if they are not supported, maybe we can just state so).

spidermon/contrib/scrapy/monitors.py
rennerocha: This setting is needed to enable the monitor, not the stats. I changed to `"To enable field coverage monitor, set SPIDERMON_ADD_FIELD_COVERAGE=True in your project settings"` so it is clearer what is the purpose of this setting.

docs/source/settings.rst
rennerocha: Thanks! It is fixed!

docs/source/settings.rst
rennerocha: They are not supported (and I don't think it is worthy to add try to add it now). I will include a note on the documentation about this.

rennerocha:
Ready for review again! :-) Hopefully this would be enough to release this feature!

docs/source/settings.rst
```diff
@@ -118,10 +118,88 @@ Default: ``[]``
...
 
-.. SPIDERMON_ADD_FIELD_COVERAGE:
-
 SPIDERMON_ADD_FIELD_COVERAGE
 ----------------------------
 Default: ``False``
 
-When enabled, Spidermon will add field coverage information in Spider stats.
+When enabled, Spidermon will add statistics about the number of items scraped and coverage for each existing
+field following the format:
```
Gallaecio, line 126: :lipstick: 

```suggestion
field following this format:
```

docs/source/settings.rst
```diff
@@ -118,10 +118,88 @@ Default: ``[]``
...
+      'spidermon_field_coverage/dict/field_1': 1,
+      'spidermon_field_coverage/dict/field_1/nested_field_1_1': 1,
+      'spidermon_field_coverage/dict/field_1/nested_field_1_2': 0.5,
+      'spidermon_item_scraped_count/dict/field_2': 0.5,
+
+SPIDERMON_FIELD_COVERAGE_SKIP_NONE
+----------------------------------
+Default: ``False``
+
+When enabled, fields returned that contains `None` as value will not be counted as an existing value.
```
Gallaecio, line 170: :lipstick: 

```suggestion
When enabled, returned fields that have ``None`` as value will not be counted as fields with a value.
```

spidermon/contrib/scrapy/monitors.py
```diff
@@ -174,6 +174,99 @@ def test_verify_item_validation_error(self):
...
+    A set of rules may be defined as follows:
+
+    .. code-block:: python
+
+        # project/settings.py
+        SPIDERMON_FIELD_COVERAGE_RULES = {
+            "dict/field_1": 0.4,  # Expected 40% coverage for field_1
+            "dict/field_2": 1.0,  # Expected 100% coverage for field_2
+            "dict/field_3": 1.0,  # Expected 100% coverage for parent field_3
+            "dict/field_3/field_3_1": 0.5,  # Expected 50% coverage for nested field_3_1
```
Gallaecio, line 211: Maybe we should explicitly indicate in the documentation below whether the value for nested fields is relative to the parent field or absolute.

tests/contrib/scrapy/test_monitors_field_coverage.py
```diff
@@ -0,0 +1,106 @@
...
+    monitor_runner = data.pop("runner")
+    try:
+        monitor_runner.run(field_coverage_monitor_suite, **data)
+    except NotConfigured:
+        pytest.fail(
+            "It should not raise NotConfigured if SPIDERMON_ADD_FIELD_COVERAGE = True"
+        )
+
+
+def test_monitor_fail_if_coverage_less_than_expected(field_coverage_monitor_suite):
```
ejulio, line 57: It seems all the tests below are the same, just changing the stats and assertion.
What about `pytest.mark.parametrize`?

```
pytest.mark.parametrize("coverage, result", [
    (0.5, False),
    (0.8, True), 
    # ...
])
def my_test(coverage, result):
    settings = {
        "SPIDERMON_ADD_FIELD_COVERAGE": True,
        "SPIDERMON_FIELD_COVERAGE_RULES": {"dict/field": 0.8,},
    }
    stats = {"spidermon_field_coverage/dict/field": coverage}
    data = make_data_for_monitor(settings=settings, stats=stats)
    monitor_runner = data.pop("runner")
    monitor_runner.run(field_coverage_monitor_suite, **data)

    assert monitor_runner.result.wasSuccessful() is result
```

spidermon/contrib/scrapy/monitors.py
rennerocha: Documentation was improved.

tests/contrib/scrapy/test_monitors_field_coverage.py
rennerocha: Not all tests are equal. One of them checks for the absence of a field. As they are just 3 tests that follows this pattern, I would prefer to keep them separated for now.

PR #265, opened 2020-08-19
Update slackclient to v2.0

andersonberg:
This PR fixes https://github.com/scrapinghub/spidermon/issues/251
Updating the slackclient to the new version.
Requirements and setup were updated to use slackclient versions higher than 2.0.
I also updated tests and docs.

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -208,13 +208,13 @@ using a bot when your monitor suites finish their execution.
 
 To use this action you need to:
 
-#.  Install `slackclient`_ 1.3 or higher, but lower than 2.0:
+#.  Install `slackclient`_ 2.0 or higher:
 
     .. code-block:: shell
 
-        $ pip install "slackclient>=1.3,<2.0"
+        $ pip install "slackclient>=2.0"
 
-    .. warning:: This action **does not** work with `slackclient`_ 2.0 or later.
+    .. warning:: This action **requires** Python 3.6 or later.
 
 #.  Provide the `Slack credentials`_ in your ``settings.py`` file as follows:
 
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -214,7 +214,6 @@ To use this action you need to:
 
         $ pip install "slackclient>=2.0"
 
-    .. warning:: This action **requires** Python 3.6 or later.
 
 #.  Provide the `Slack credentials`_ in your ``settings.py`` file as follows:
 
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -198,6 +198,8 @@ Slack notifications
 Here we will configure a built-in Spidermon action that sends a pre-defined message to
 a Slack channel using a bot when a monitor fails.
 
+  .. warning:: Slack integration **requires** Python 3.6 or later.
+
 .. code-block:: python
 
     # tutorial/monitors.py
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -198,8 +198,6 @@ Slack notifications
 Here we will configure a built-in Spidermon action that sends a pre-defined message to
 a Slack channel using a bot when a monitor fails.
 
-  .. warning:: Slack integration **requires** Python 3.6 or later.
-
 .. code-block:: python
 
     # tutorial/monitors.py
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 six
 Jinja2
-slackclient>=1.3.0,<2.0.0
+slackclient>=2.0.0
 boto
 premailer
 jsonschema[format]
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 six
 Jinja2
-slackclient>=2.0.0
+slackclient
 boto
 premailer
 jsonschema[format]
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ test_requirements = [
     "lxml",
     "premailer",
     "scrapy",
-    "slackclient>=1.3.0,<2.0.0",
+    "slackclient>=2.0.0",
     "twisted>=19.7.0",
 ]
 
@@ -32,7 +32,7 @@ setup(
         "monitoring": [
             "scrapy",
             "Jinja2",
-            "slackclient>=1.3.0,<2.0.0",
+            "slackclient>=2.0.0",
             "boto",
             "premailer",
             "sentry-sdk",
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ test_requirements = [
     "lxml",
     "premailer",
     "scrapy",
-    "slackclient>=2.0.0",
+    "slackclient",
     "twisted>=19.7.0",
 ]
 
@@ -32,7 +32,7 @@ setup(
         "monitoring": [
             "scrapy",
             "Jinja2",
-            "slackclient>=2.0.0",
+            "slackclient",
             "boto",
             "premailer",
             "sentry-sdk",
--- a/spidermon/contrib/actions/slack/__init__.py
+++ b/spidermon/contrib/actions/slack/__init__.py
@@ -6,7 +6,6 @@ import logging
 
 import six
 from slack import WebClient
-from slack.errors import SlackApiError
 
 from spidermon.contrib.actions.templates import ActionWithTemplates
 from spidermon.exceptions import NotConfigured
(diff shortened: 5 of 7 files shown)


spidermon/contrib/actions/slack/__init__.py
```diff
@@ -105,7 +105,9 @@ def _api_call(self, method, **kwargs):
         return response
```
rennerocha, line 105: **suggestion** This line added some validation that was required by older versions of `slackclient`. Maybe it may be removed (or updated):

```if isinstance(response, six.string_types):  # slackclient < v1.0```

File: spidermon/contrib/actions/slack/__init__.py [hunk 2]
```diff
@@ -92,8 +92,8 @@ def _get_users_info(self):
             ]
         )
 
-    def _api_call(self, method, **kwargs):
-        response = self._client.api_call(method, **kwargs)
+    def _api_call(self, method, json=None):
+        response = self._client.api_call(method, json=json)
```
rennerocha, line 96: **suggestion**: According to [migration guide](https://github.com/slackapi/python-slackclient/wiki/Migrating-to-2.x#web-client-api-changes) in the example code, the use of `api_call` method is still allowed, but the most efficient way (as they mentioned) is to use `chat_postMessage` call. Perhaps it would be better to stop using `api_call` and using it for all requests we make (I think it would affect calls to `conversations.open` and `users.list`) following the library recommendation.

spidermon/contrib/actions/slack/__init__.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, sender_token=None, sender_name=None, fake=False):
             raise NotConfigured("You must provide a slack sender name.")
 
         self.fake = fake
-        self._client = SlackClient(sender_token)
+        self._client = WebClient(sender_token)
         self._users = None
 
     @property
```
rennerocha, line 33: **thought**: We have [this PR](https://github.com/scrapinghub/spidermon/pull/260/) opened that changes error handling for `_get_users_info`. If we are going to change how the call is made (as I commented some lines below), maybe we should try to incorporate her changes to make sure that we will not have the issue or, if the code changes are too much, take care to solve the problem she mentioned.

setup.py
```diff
@@ -8,7 +8,7 @@
     "lxml",
     "premailer",
     "scrapy",
-    "slackclient>=1.3.0,<2.0.0",
+    "slackclient>=2.0.0",
```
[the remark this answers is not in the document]
Gallaecio: Shouldn’t we wait until 3.5 support is dropped to merge this?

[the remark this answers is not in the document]
rennerocha: I agree with that. Just waiting for https://github.com/scrapinghub/spidermon/issues/266 to be done, than we won't need to change anything here.

spidermon/contrib/actions/slack/__init__.py
andersonberg: It doesn't seem to me that the changes will be too much... We can choose to go either of the two ways: I can incorporate her changes to this PR, or after this PR is merged, @karenyavine can update her PR. What do you think?

docs/source/actions.rst
```diff
@@ -208,13 +208,13 @@ using a bot when your monitor suites finish their execution.
...
-#.  Install `slackclient`_ 1.3 or higher, but lower than 2.0:
+#.  Install `slackclient`_ 2.0 or higher:
 
     .. code-block:: shell
 
-        $ pip install "slackclient>=1.3,<2.0"
+        $ pip install "slackclient>=2.0"
 
-    .. warning:: This action **does not** work with `slackclient`_ 2.0 or later.
+    .. warning:: This action **requires** Python 3.6 or later.
```
rennerocha, line 217: **suggestion (non-blocking)** As Python 3.5 support was dropped, there is no need to include this warning on the docs now.

docs/source/getting-started.rst
```diff
@@ -198,6 +198,8 @@ Slack notifications
 Here we will configure a built-in Spidermon action that sends a pre-defined message to
 a Slack channel using a bot when a monitor fails.
 
+  .. warning:: Slack integration **requires** Python 3.6 or later.
```
rennerocha, line 201: **suggestion (non-blocking)** As Python 3.5 support was dropped, there is no need to include this warning on the docs now.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 six
 Jinja2
-slackclient>=1.3.0,<2.0.0
+slackclient>=2.0.0
```
rennerocha, line 3: **suggestion (non-blocking)** I believe that we don't need to specify a minimum version here, as the latest version of this library should work correctly now.

setup.py
```diff
@@ -32,7 +32,7 @@
         "monitoring": [
             "scrapy",
             "Jinja2",
-            "slackclient>=1.3.0,<2.0.0",
+            "slackclient>=2.0.0",
```
rennerocha, line 35: **suggestion (non-blocking)** You got the idea :-)

tests/contrib/actions/slack/test_slack_action.py
```diff
@@ -8,6 +9,7 @@ def logger_info(mocker):
     return mocker.patch("spidermon.contrib.actions.slack.logger.info")
 
 
+@pytest.mark.skipif(sys.version_info < (3, 6), reason="requires python3.6 or higher")
```
rennerocha, line 12: **suggestion**: Python 3.5 support was dropped. This can be removed now.

spidermon/contrib/actions/slack/__init__.py
```diff
@@ -85,27 +86,26 @@ def _get_user_id(self, username):
...
-
-        has_errors = not response.get("ok")
-        if has_errors:
-            error_msg = response.get("error", {}).get("msg", "Slack API error")
-            logger.error(error_msg)
-
-        if isinstance(response, six.string_types):  # slackclient < v1.0
-            response = json.loads(response)
-        return response
+        try:
```
rennerocha, line 89: **suggestion** Considering that we don't do anything special (like retrying the call or other action to recover from the exception) and just logging the exception text, I believe that removing the `try...except` is fine as the exceptions will be logged anyway. The same for all calls below.

rennerocha (changes requested):
@andersonberg 
With the removal of Python 3.5 support, some changes are not needed anymore. Also I added a few suggestion on other changes you made.

spidermon/contrib/actions/slack/__init__.py
andersonberg: I made some tests using the changes suggested by that PR. Those changes only work when Spidermon sends messages to a channel, but it will fail when sending messages to a single user (private messages), though. That's because to send a message we need the `user_id` that comes from the `users_list` method.

spidermon/contrib/actions/slack/__init__.py
andersonberg: I'm adding the `try..except` blocks because of this warning: https://github.com/slackapi/python-slackclient/wiki/Migrating-to-2.x#error-handling-changes
Will Spidermon always log those exception even if we don't add them to `logger.error`? (sorry if it's an obvious question)

spidermon/contrib/actions/slack/__init__.py
karenyavine: I hadn't considered the bot messaging via DM (is it even used? I'd assume alerting via dm would be less popular?). But It think it’d be safest to ask for the `user_id` to be supplied as an config, in order to maintain the least amount of privs.

andersonberg: It would be the same for a single user or a list of users. In that case, the Spidermon settings would only accept user_id(s) and not username(s) anymore. I agree the safest way is to use the `users_info` method instead of the `users_list`, but I can't see another solution

rennerocha: I did some research on this topic and maybe we won't have a better solution other than iterating over the entire list of users until we find the one we want.

We allow messaging via DM and as we can define a list of users that will receive the message, this is a feature that we should keep providing. I looked at the internal codebase of Scrapinghub and found the the majority uses a channel, but there are a few using DM.

According to [Slack](https://api.slack.com/changelog/2017-09-the-one-about-usernames), they will not provide an easy way to get `user_id` base on the username (they want to get rid of usernames!), so the only solution for now will be continue [iterating over all users](https://github.com/slackapi/node-slack-sdk/issues/73#issuecomment-497127404) until we find the one we want. To be fair, this makes no sense to me!

Asking for the `user_id` in the config (instead of the username) will add an extra burden to the user, because it is not that easy (not programatically) to discover the `user_id` of a user.

spidermon/contrib/actions/slack/__init__.py
rennerocha: Any exception raised will be sent to stderr, so we don't need to care about it.

spidermon/contrib/actions/slack/__init__.py
```diff
@@ -85,27 +86,26 @@ def _get_user_id(self, username):
...
-        return self._api_call("im.open", user=user_id)["channel"]["id"]
+        try:
+            response = self._client.conversations_open(users=[user_id])
+            return response["channel"]["id"]
+        except SlackApiError as e:
+            assert e.response["ok"] is False
+            assert e.response["error"]
+            logger.error(e.response["error"])
 
     def _send_user_message(
```
rennerocha, line 110: [user_channel](https://github.com/scrapinghub/spidermon/pull/265/files#diff-56dc9c5d7e8074bc42c77ec35c9c09c2L117) needs to be the same as `user_id`, or else the message will not be sent to the user.

rennerocha (changes requested):
DM is not working due the lack of the correct channel value.

PR #268, opened 2020-09-04
Fix code formatting with black

rennerocha:
Build was failing because code wasn't formatted accordingly using black.


rennerocha:
This PR doesn't need more tests as it is only updating code formatting. For some unknown reason, coverage starts to consider one line as not covered even if the same line formatted in a different way was considered as covered. So I will merge this PR as it is now.

PR #271, opened 2020-09-10
Test on Python 3.9-dev and 3.10-dev

hugovk:
Python 3.9.0 is due out next month, it's a good idea to test the release candidate already, to make sure both this library and Python itself are working.

* https://www.python.org/dev/peps/pep-0596/#release-schedule

Optionally 3.10-dev can be tested too, although it's not yet in alpha.

* https://www.python.org/dev/peps/pep-0619/#release-schedule

Also remove `sudo: false`, it no longer does anything: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 git:
   depth: false
   quiet: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ python:
   - 3.6
   - 3.7
   - 3.8
+  - 3.9-dev
+  - 3.10-dev
 install:
   - pip install tox-travis codecov
 script:


Gallaecio (approved):
Good thinking!

rennerocha:
Testing in newer release may be a good idea. However, if tests fail in one of these versions, the build will break which is not good (this prevents a PR to be merged). This is unlikely to happen, but we had small issues when we introduce Python 3.8 to the project, so I tend to be against introduce this check in our pipeline.

When a newer Python version is released, we should create a specific PR to make sure it work. With luck it will be just a matter of adding the version number in travis and update the docs.

hugovk:
You can always pop it behind `allow_failures` to prevent failing the build whilst it's investigated:

* https://docs.travis-ci.com/user/build-matrix/#rows-that-are-allowed-to-fail

---

The Python 3.9 release manager said of the newest release candidate:

> In the mean time, we **strongly encourage** maintainers of third-party Python projects to prepare their projects for 3.9 compatibility during this phase. As always, report any issues to [the Python bug tracker](https://bugs.python.org/).

https://discuss.python.org/t/python-3-9-0rc2-is-now-available-for-testing/5236

RC builds should be stable. If something fails, it's probably either something in the library that needs fixing (e.g. API removals like https://github.com/scrapinghub/spidermon/pull/272 for 3.9, or as you mention, the 3.8 API change in https://github.com/scrapinghub/spidermon/pull/264/commits/ba21618e49f22d8f22e01239d0e60b4bdc173955) and is best fixed before the Python release. Fail fast: catch it in the CI months before 3.x is released and end users trip over.

Or it's something in CPython itself, and you're helping the community test CPython itself, to help ensure a high quality release.

Anyway, 3.9.0 final is due out on Monday, so at this stage you may as well wait until it's available on Travis in a week or so.

---

Python 3.10 is still bleeding edge (alpha due on Monday too), so it's much less important to test this one than RCs.

PR #279, opened 2020-12-30
Sentry message improvement

wRAR:
As discussed with Renne before, this is a small batch of improvements for `SendSentryMessage` copied from internal SH projects and already used there for some time. I've split them into several commits:

1. Add the spider name to tags.
2. Add the failed monitor names to tags.
3. Calculate a hash of the failed monitor names and add it to tags.

I omitted another thing: setting the fingerprint to `[spider_name, reason_hash]` (so that the reports are only grouped into one if the list of failed monitors matches) as that's backward incompatible and probably needs either a setting or an easy way to do this in a subclass.


spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -100,15 +102,47 @@ def get_message(self):
...
+                try:
+                    k = mon[1].lower().replace(" ", "_")
+                    # 32 chars undocumented tag key length
+                    tags.update({k[:32]: 1})
+                except IndexError:
+                    pass
+        return tags
+
+    @staticmethod
+    def get_reason_hash(message):
```
rennerocha, line 121: **suggestion**: Albeit this tag can help to group notification with the same failing monitors, maybe this is not general enough and could be better accomplished as a custom action inherited from this action.

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -100,15 +102,47 @@ def get_message(self):
...
+    @staticmethod
+    def get_tags(message):
+        tags = {}
+        failed_monitors = message.get("failed_monitors", [])
+        failed_monitors = [y.split("/") for y in failed_monitors]
+        for key, group in groupby(failed_monitors, key=lambda x: x[0]):
+            for mon in group:
+                try:
+                    k = mon[1].lower().replace(" ", "_")
+                    # 32 chars undocumented tag key length
```
rennerocha, line 114: **question**: Is this a limit from Sentry (not allowing tags with more than 32 chars)? If so, maybe it is worthy to include that information in the documentation of the action.

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -100,15 +102,47 @@ def get_message(self):
...
 
+    @staticmethod
+    def get_tags(message):
+        tags = {}
+        failed_monitors = message.get("failed_monitors", [])
+        failed_monitors = [y.split("/") for y in failed_monitors]
+        for key, group in groupby(failed_monitors, key=lambda x: x[0]):
+            for mon in group:
+                try:
+                    k = mon[1].lower().replace(" ", "_")
```
rennerocha, line 113: **issue**: When failed method name has special characters (like `á`, `*`, etc), the tag is not applied correctly and Sentry adds an error in the notification. 

invalid_chars

**suggestion**: It would be great if we handle these specials characters to avoid this problem.

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -100,15 +102,47 @@ def get_message(self):
...
+                    # 32 chars undocumented tag key length
+                    tags.update({k[:32]: 1})
```
rennerocha, line 115: **question**: What if I have two methods with the same name (at least in the first 32 characters)?

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -100,15 +102,47 @@ def get_message(self):
 
         return message
 
+    @staticmethod
+    def get_tags(message):
```
rennerocha, line 106: **suggestion**: We don't have any tests for this action. It would be good to have at least some tests for `get_tags` method.

rennerocha (changes requested):
Added a few comments and questions on the code. Please let me know if you have any questions.

**suggestion**: Include documentation explaining what tags are added to the notification.

spidermon/contrib/actions/sentry/__init__.py
wRAR: I don't think we have tests for anything in this module, what do you suggest?

spidermon/contrib/actions/sentry/__init__.py
wRAR: Actually, it's now [documented](https://develop.sentry.dev/sdk/data-handling/):

```
Tag keys are limited to 32 characters.
Tag values are limited to 200 characters.
```

spidermon/contrib/actions/sentry/__init__.py
wRAR: Yeah, there will be only one tag then. Note that this uses human-readable names if available so there is probably more variety. I also wonder if should strip leading "test_" to save some space.

spidermon/contrib/actions/sentry/__init__.py
wRAR: Ouch. I wonder how should we escape them (dropping them looks bad).

spidermon/contrib/actions/sentry/__init__.py
wRAR: Yeah, no problem with that. There are also some other ways to configure merging documented at https://docs.sentry.io/product/sentry-basics/guides/grouping-and-fingerprints/

spidermon/contrib/actions/sentry/__init__.py
```diff
@@ -100,15 +102,47 @@ def get_message(self):
...
+            tags = self.get_tags(message)
+            for key, val in tags.items():
+                scope.set_tag(key, val)
+
+            reason_hash = self.get_reason_hash(message)
+            scope.set_tag("reason_hash", reason_hash)
+
+            spider_name = message.get("spider_name", "")
+
+            scope.set_tag("spider_name", spider_name)
```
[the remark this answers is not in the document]
wRAR: Sounds good.

spidermon/contrib/actions/sentry/__init__.py
rennerocha: **suggestion**: If you decorate your monitor class and/or your monitor methods  with `@monitors.name`, the `test_` part will not be added, so it won't avoid duplication. I think it is worthwhile to add a warning about this limitation in [Sentry action](https://spidermon.readthedocs.io/en/latest/actions.html#sentry-action) documentation and the user can take care of not using long names for its monitor.

**suggestion**: Given that the format of the `failed_monitors` will always be <ClassName>/<MethodName>` (even if the names are defined by the user) I think `get_tags` could be simplified to:

```
    def get_tags(self, message):
        tags = {
            "spider_name": message.get("spider_name", ""),
            "project_name": self.project_name,
        }

        for failed_monitor in message.get("failed_monitors", []):
            # 32 chars limited by Sentry
            key = slugify(failed_monitor.split("/")[-1])[:32]
            tags[key] = 1

        return tags
```

spidermon/contrib/actions/sentry/__init__.py
rennerocha: **suggestion**: Maybe we may consider using a library like [python-slugify](https://github.com/un33k/python-slugify) to generate proper slug for this string, removing/replacing unwanted characters. This library is already a spidermon dependency (for [validation stats](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/scrapy/stats.py#L2)), so you can use it with no need to include an extra requirement to the project.

rennerocha (changes requested):
Almost good to merge, just minor changes to make it work in a more general way):
- Add slugify to generate better names for the tags (excluding unicode invalid values)
- Document `get_tags` method informing which tags are included to the notification and adding an warning about the 32 chars limits for the tag name.

rennerocha (approved):

PR #284, opened 2021-03-22
Add some more Scrapy monitors

wRAR:
This PR adds 5 new monitors:

* WarningCountMonitor
* RetryCountMonitor
* DownloaderExceptionMonitor
* SuccessfulRequestsMonitor
* TotalRequestsMonitor

I didn't add them to `SpiderCloseMonitorSuite` but maybe we should (their defaults should be reasonable).


wRAR:
Reviewers should check that all names and comments are correct, as there was a lot of copy-pasting :)

tests/test_monitors.py
```diff
@@ -138,6 +148,55 @@ def test_log_monitor_should_pass(make_data):
     assert runner.result.monitor_results[0].error is None
 
 
+def test_warning_monitor_should_fail(make_data):
+    """ WarningCount should fail if the # of warning log messages exceed the limit """
+
+    # Scenario # 1
```
rennerocha, line 154: **issue**: It is not a good practice to have more than one scenario being tested in the same method. This would make difficult to understand what scenario failed in the test. It will be better to have a separated method for each scenario. It also needs one test for each possible situation (disabled monitor with -1, failing and passing).

tests/test_monitors.py
```diff
@@ -138,6 +148,55 @@ def test_log_monitor_should_pass(make_data):
...
+    # Scenario # 2
+    data = make_data({SPIDERMON_MAX_WARNINGS: 10})
+    runner = data.pop("runner")
+    suite = new_suite([WarningCountMonitor])
+    data["stats"]["log_count/WARNING"] = 12
+    runner.run(suite, **data)
+    assert "Found 12 warnings in log" in runner.result.monitor_results[0].error
+
+
+def test_warning_monitor_should_pass(make_data):
```
rennerocha, line 171: **issue**: This tests if the monitor doesn't fail if disabled. Comment is wrong.

tests/test_monitors.py
```diff
@@ -7,12 +7,22 @@
     FinishReasonMonitor,
     ItemCountMonitor,
     ErrorCountMonitor,
+    WarningCountMonitor,
```
rennerocha, line 10: **suggestion (non-blocking)**: This test module is becoming too long and it is difficult to find specific tests for specific monitors here. Maybe we could reorganize these tests so we have one test module for each monitor we are testing, so we can find the tests we need more easily. This certainly can be made in a different PR.

My initial idea would be replace this `test_monitors.py` file with several files organized like:
```
/tests
  /contrib
    /scrapy
      /monitors
        /test_warning_count_monitor.py
        /test_successful_requests_monitor.py
        /test_total_requests_monitor.py
        (etc etc)
```

rennerocha:
>     * RetryCountMonitor

All monitors look good. Nothing against adding them to `SpiderCloseMonitorSuite`. It would be good to add a little more explanation in the docs that if you want just a set of these monitors, you need to create your monitor suite and configure it with the monitors you want. I think [this section of docs](https://spidermon.readthedocs.io/en/latest/monitors.html#is-there-a-basic-scrapy-suite-ready-to-use) a good place for that.

rennerocha (changes requested):
All these monitors are very useful (I am working on projects that may benefit from them!). Just left a few comments about the tests and some improvements to the docs and we should be good.

rennerocha:
@wRAR now I noticed that the issue on the build happened on Travis CI. We removed Travis in favor of Github Action [three months ago](https://github.com/scrapinghub/spidermon/commit/21efc0b3177fd4c887a1fc5154bc255e3ee8a32d). It appears that your PR stayed without reviews for that long! Sorry about that! Could you please rebase your branch with `master` branch and check if everything is working correctly (I don't see any reason for not)? Probably after that, the build will pass and we will be able to merge it. I see no other issues that may prevent it to be merged now.

wRAR:
I'm waiting for #297 to be merged to be able to split this one too.

rennerocha:
@wRAR #297 is merged! You should be good to rebase it and update this branch now!

PR #286, opened 2021-04-06
spidermon.contrib.actions.email.ses: switch to boto3

### LINKED ISSUE #285: Email has stopped working
[It looks like we need to switch to boto3](https://stackoverflow.com/a/63454370/939364) to fix it.

Gallaecio:
Fixes #285

Based on https://stackoverflow.com/a/52105406/939364

I still need to test this (manually).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
             "Jinja2",
             "slackclient",
             "boto",
+            "boto3",
             "premailer",
             "sentry-sdk",
         ],
--- a/spidermon/contrib/actions/email/ses.py
+++ b/spidermon/contrib/actions/email/ses.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
-from boto import ses
+
+import boto3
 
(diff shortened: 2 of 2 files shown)


Gallaecio:
Manual test succeeded, automatic tests passed (the Python 3.6 job failure is unrelated), and rebased as a single commit.

rennerocha:
Interestingly, if you still use `boto`, the email is sent. I use one credentials set (SPIDERMON_AWS_ACCESS_KEY and SPIDERMON_AWS_SECRET_KEY) and the message was delivered correctly. Perhaps the problem happens only with specific accounts (created after a certain date, or with some set of permissions).

In any case, using `boto3` is the [recommended](https://github.com/boto/boto#boto-3) library to be used, so we need to start using it here.

rennerocha (approved):

Gallaecio:
From https://stackoverflow.com/a/63454370/939364 I gather that they are making the change progressively, not to all customers at once.

PR #291, opened 2021-04-17
Add Github Actions support and remove Travis

rennerocha:


Gallaecio (approved):
It looks good to me.

It seems cleaner than Scrapy’s in many aspects (e.g. publishing on release creation seems nice).

It may make sense to extract the static checks (e.g. doc build) into a separate workflow file, like Scrapy does, so that they can run in parallel to the rest instead of at the end of the tests that use the latest Python version. But I don’t see a problem with merging as is for now, we can always change that in the future (e.g. if more static checks are added).

PR #297, opened 2021-06-25
Split tests/test_monitors.py into per-monitor files.

wRAR:
As suggested in #284 I've split the monolithic test_monitors.py. I've noticed that there is a pattern of `MonitorSuite(monitors=[FooMonitor])` that is implemented as a function in 3 cases but as a fixture in 1 case, I think these usages should be made consistent. 

This doesn't split multiple-scenario tests for UnwantedHTTPCodesMonitor, I'm fine with doing this here if suggested.


wRAR:
The `black` test suggests reformatting two affected and one unaffected file (looks like it's different between 21.6b0 and 20.8b1).

rennerocha:
> The `black` test suggests reformatting two affected and one unaffected file (looks like it's different between 21.6b0 and 20.8b1).

black 21.5b2 stopped to add an empty space in docstrings. Could you please apply black on your code using the latest version?

rennerocha (approved):

Issue #306, opened 2021-09-06
Fix permissions based errors when sending slack messages


further-reading:
The current implementation of slack messaging causes errors when certain permissions are missing. The issue come from [the following line](https://github.com/scrapinghub/spidermon/blob/94d4c3291925803c86ef5e892a2fdc228697a20a/spidermon/contrib/actions/slack/__init__.py#L118)

The icon_url argument is an optional argument when sending a slack message to change the avatar displayed when the message is sent. In order to find the content to place in this argument the code takes two steps  
1. Gets a dict that contains details of all users in a slack organisation.  
2. Searches for the `SPIDERMON_SLACK_SENDER_NAME` in in this dict.  

However, these steps are prone to fail. First, to achieve step 1 the `users:read` permission needs to be added to the bot, but a bot that is intended to just write messages to slack should function with only the `chat:write` permission. Second, it is possible for some integrations to not appear in the specific list that spidermon is looking for users in, resulting in a key error when trying to grab its own details.

As I mentioned this is an optional argument - if this is not included in the slack `chat_postMessage` call then it just uses the icon that was set by the bot's developer. The only use case is when the bot is using an icon from another user/app within the organisation, which feels like it's probably an edge case. As such I propose adding error handling when attempting to find the icon_url so basic slack bots will work as expected.

PR #307, opened 2021-09-06
[SlackMessageManager] Adding error handling for common permission issue

further-reading:
fixes https://github.com/scrapinghub/spidermon/issues/306

The current implementation of slack messaging has the bot explicitly set its avatar when sending the message. This step is usually unnecessary - it is only required if the sender_name is different to the bot's name as if this argument is left blank it will use the same icon that was added when creating the bot. The method it uses to find the icon_url will fail if it's not set up in a specific way, which adds additional permission requirements and complexity when integrating slack messaging.

This change will allow the bot to still send a message if it can't find an icon_url.


further-reading:
I'm not sure why those tests are failing - my code doesn't touch anything related to them

spidermon/contrib/actions/slack/__init__.py
```diff
@@ -115,9 +116,36 @@ def _send_channel_message(
...
+        try:
+            icon_url = self.users[self.sender_name]["profile"]["image_48"]
+        except SlackApiError as e:
+            if (
+                e.response.data.get("error") == "missing_scope"
+                and e.response.data.get("needed") == "users:read"
+            ):
+                # bot does not have read permissions for slack org
+                # can be an expected outcome - will use its own icon
+                icon_url = None
```
rennerocha, line 140: **suggestion (non-blocking)** I understood that this is not en error as it is a completely valid scenario for a bot, but maybe it would be useful to add a warning log when these exceptions are raised, just for the records in the case that the user really wants another icon to be used.

spidermon/contrib/actions/slack/__init__.py
```diff
@@ -115,9 +116,36 @@ def _send_channel_message(
...
+            else:
+                raise e
+        except KeyError:
+            # bot has read permissions for slack org but can't find sender in list
+            # can be an expected outcome - will use its own icon
+            icon_url = None
```
rennerocha, line 146: **suggestion (non-blocking)** Add a warning here as well for the same reason as my previous comment.

rennerocha (approved):
@further-reading code looks good and the solution is really very straightforward. I added just one non-blocking suggestion. Also, could you please rebase your branch with master (that fixed the tests), so we can have a new passing build?

further-reading:
Thanks @rennerocha - updated now + also added updates to the slack bot setup guide about setting OAuth permissions.

Issue #300, opened 2021-07-21
Built-in monitor for spider running time


rennerocha:
When we have long running jobs, usually is useful to be notified if the spider is running for more time than we expect. This may indicate some problem in the spider (infinite redirection), lots of retries or other problem.

A monitor that checks the running time [periodically](https://spidermon.readthedocs.io/en/latest/monitors.html#periodic-monitors) against a `SPIDERMON_MAX_EXECUTION_TIME` setting would be a good improvement.

PR #309, opened 2021-09-11
[PeriodicExecutionTimeMonitor] Creating monitor for runtime exceeding a target maximum

further-reading:
fixes https://github.com/scrapinghub/spidermon/issues/300


spidermon/contrib/scrapy/monitors.py
```diff
@@ -275,6 +278,30 @@ def test_check_if_field_coverage_rules_are_met(self):
...
 
+@monitors.name('Periodic execution time monitor')
+class PeriodicExecutionTimeMonitor(Monitor, StatsMonitorMixin):
+    """Check for runtime exceeding a target maximum runtime.
+
+        You can configure the maximum runtime (in seconds) using
+        ``SPIDERMON_MAX_EXECUTION_TIME``."""
+
+    @monitors.name('Maximum execution time reached')
+    def test_runtime(self):
```
rennerocha, line 289: **nitpick** `test_execution_time`

spidermon/contrib/scrapy/monitors.py
```diff
@@ -300,3 +327,17 @@ class SpiderCloseMonitorSuite(MonitorSuite):
...
         FieldCoverageMonitor,
     ]
+
+
+class PeriodicMonitorSuite(MonitorSuite):
+    """This Monitor Suite implements the following monitors:
+
+        * :class:`PeriodicExecutionTimeMonitor`
+
+        You can easily enable this monitor *after* enabling Spidermon::
```
rennerocha, line 337: **question** What you mean with *after* enabling Spidermon?

rennerocha:
@further-reading looks good. Need some unit tests.

spidermon/contrib/scrapy/monitors.py
further-reading: I copied the docstring from SpiderCloseMonitorSuite. My understanding is that it's explaining that spidermon must be enabled for these standard monitors to be used.

SciPie (approved):
Looks good.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -495,6 +498,31 @@ def test_check_if_field_coverage_rules_are_met(self):
...
+    ``SPIDERMON_MAX_EXECUTION_TIME`` as a project setting or spider attribute."""
+
+    @monitors.name("Maximum execution time reached")
+    def test_execution_time(self):
+        crawler = self.data.get("crawler")
+        max_execution_time = crawler.settings.getint(SPIDERMON_MAX_EXECUTION_TIME)
+        if not max_execution_time:
+            return
+        now = datetime.datetime.now()
+        start_time = self.data.stats.get("start_time")
```
curita, line 515: **question:** I believe start_time is already a `datetime` object? At least that's what it looks like in the stats when running locally.
```
2022-01-04 14:22:34 [scrapy.statscollectors] INFO: Dumping Scrapy stats:
{'elapsed_time_seconds': 0.015785,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2022, 1, 4, 17, 22, 34, 348831),
 'log_count/DEBUG': 1,
 'log_count/ERROR': 1,
 'log_count/INFO': 11,
 'memusage/max': 134189056,
 'memusage/startup': 134189056,
 'start_time': datetime.datetime(2022, 1, 4, 17, 22, 34, 333046)}
```

spidermon/contrib/scrapy/monitors.py
```diff
@@ -495,6 +498,31 @@ def test_check_if_field_coverage_rules_are_met(self):
...
+    You can configure the maximum runtime (in seconds) using
+    ``SPIDERMON_MAX_EXECUTION_TIME`` as a project setting or spider attribute."""
+
+    @monitors.name("Maximum execution time reached")
+    def test_execution_time(self):
+        crawler = self.data.get("crawler")
+        max_execution_time = crawler.settings.getint(SPIDERMON_MAX_EXECUTION_TIME)
+        if not max_execution_time:
+            return
+        now = datetime.datetime.now()
```
curita, line 514: **suggestion:** I think this might need to be `datetime.datetime.utcnow()` as I think that's what it's used to generate start_time.

curita (commented):
Had a quick glance at this because we implemented something similar at Netflix recently, and noticed a couple of things that weren't quite in line with what I think was expected on those stats.

It's merged now but it might be good to test this with an actual Scrapy job to see if the alert is triggered as expected.

spidermon/contrib/scrapy/monitors.py
further-reading: > **question:** I believe start_time is already a `datetime` object? At least that's what it looks like in the stats when running locally.

Ah yes, the zyte dash stats tab displays it as a unix timestamp but I can see that it was converted from a datetime when the spider was finished. Thanks for calling this out!

Issue #302, opened 2021-07-21
Built-in monitor for max number of failed responses allowed


rennerocha:
Validate maximum amount of `downloader/response_status_count/XXX` for failed responses (4xx or 5xx responses) in general. Maybe have two configurations, one for client errors (4xx) and other for server errors (5xx).

If we want a more granular control (for example, set a threshold only for 503 errors), we already have [UnwantedHTTPCodesMonitor](https://spidermon.readthedocs.io/en/latest/monitors.html#spidermon.contrib.scrapy.monitors.UnwantedHTTPCodesMonitor)

arslan-shafiq-zyte:
Hi @rennerocha 
I am looking into it. Thank you.

arslan-shafiq-zyte:
Hi @rennerocha 
I am looking into it. Thank you.

PR #311, opened 2021-09-14
Add UnwantedHTTPCodesFamilyMonitor Scrapy Monitor

shafiq-muhammad:
Added **UnwantedHTTPCodesFamilyMonitor** monitor to spidermon


rennerocha:
This spider is closely related to the existing `UnwantedHTTPCodesMonitor` . I will look with more care, maybe we can use it instead of creating a new one.

curita:
We have talked with Renne about this and it looks like it might be better to stick with the functionality of the existing one, as declaring the specific allowed http error codes is more explicit than declaring whole groups of allowed error codes.

@rennerocha Maybe it would be better to redefine and/or close the idea from https://github.com/scrapinghub/spidermon/issues/302 (which this PR implements) to get more clarity on the feature request.

rennerocha:
@shafiq-muhammad thanks for the PR. Sorry for the delay to review it. I checked how the proposed monitor works and it has the same goal as the [UnwantedHTTPCodesMonitor](https://spidermon.readthedocs.io/en/latest/monitors.html#spidermon.contrib.scrapy.monitors.UnwantedHTTPCodesMonitor) monitor that already exists in Spidermon. The main difference is that the existing monitor ask explicitly which are the codes that we want to monitor.

Explicit is better than implicit (as Zen of Python tells), so I think that asking for the threshold for specific status codes are more useful as a built-in monitor. If a use case demands another kind of thresholds, I think it will work better as a custom project-specific monitor.

Issue #315, opened 2021-10-03
Remove use of private module from jsonschema


rennerocha:
`jsonschema` latest version removed `jsonschema.compat` module. [According to its author]([https://github.com/Julian/jsonschema/issues/843#issuecomment-931247200), this module was intended to be private and not used by external users.

We are using this module in some JSON Schema [custom validators](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/validation/jsonschema/formats.py#L2), and I received reports from users that had problems with it.

Considering that `jsonschema.compat` was created just for compatibility with older Python version, and that we don't have this requirement anymore, we should update there to avoid this issue when using the newest version of the library.

aaneto:
Hi @rennerocha!

I have started a local fork of the repository and would like to work on this issue :)

aaneto:
Hi,

I'm closing this as this issue was fixed in #364 :)

PR #316, opened 2021-10-03
WIP: Fix compatibility issues with jsonschema>=4

cuducos:
The only thing we used from this module was the `str_types` [which, for Python 3, is just `str`](https://github.com/Julian/jsonschema/blob/077feca5c176dd740b30d3f1084b6622076129e7/jsonschema/compat.py#L26). In the source it's a tuple, but I  `isinstance` accepts both a type/class or a tuple, so I kept it simple.

Addresses #315.

### DIFF
--- a/tests/test_validators_jsonschema.py
+++ b/tests/test_validators_jsonschema.py
@@ -43,13 +43,21 @@ class AdditionalItems(SchemaTest):
     data_tests = [
         DataTest(
             name="additionalItems as schema, additional items match schema",
-            schema={"items": [{}], "additionalItems": {"type": "integer"}},
+            schema={
+                "$schema": "http://json-schema.org/draft-07/schema#",
+                "items": [{}],
+                "additionalItems": {"type": "integer"},
+            },
             data=[None, 2, 3, 4],
             valid=True,
         ),
         DataTest(
             name="additionalItems as schema, additional items do not match schema",
-            schema={"items": [{}], "additionalItems": {"type": "integer"}},
+            schema={
+                "$schema": "http://json-schema.org/draft-07/schema#",
+                "items": [{}],
+                "additionalItems": {"type": "integer"},
+            },
             data=[None, 2, 3, "foo"],
             valid=False,
             expected_errors={"3": [messages.INVALID_INT]},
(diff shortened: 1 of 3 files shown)


cuducos:
Just a bit of clarification: this PR passes the tests with `jsonschema==3.2.0` (for example, adding `deps = jsonschema==3.2.0` in `tox.ini` under `[testenv]`).

If we update `jsonschema` to the newest version, 4.0.1, it looks like there are more incompatibilities.

I'm not sure if maintainers here prefer more atomic PRs or if we should address further incompatibilities with `jsonschama==4.0.1` here. Please, let me know.

rennerocha:
I released a new version of Spidermon (1.15.2) with `jsonschema` library pinned to the latest stable version, so you can use it now in your projects without worrying about the dependencies versions.

For this specific issue, related to the newest version of `jsonschema`, I'd prefer a bigger PR that address all problems that was created by version 4.x of `jsonschema` if possible.

cuducos:
> For this specific issue, related to the newest version of `jsonschema`, I'd prefer a bigger PR that address all problems that was created by version 4.x of `jsonschema` if possible.

Good to know. So I'm gonna rename this PR and continue to work on it!

cuducos:
I pushed a WIP (_work in progress_) commit to check with maintainers some doubts about what is desired for this package. Please, follow the inline comments ; )

tests/test_validators_jsonschema.py
```diff
@@ -43,13 +43,21 @@ class AdditionalItems(SchemaTest):
     data_tests = [
         DataTest(
             name="additionalItems as schema, additional items match schema",
-            schema={"items": [{}], "additionalItems": {"type": "integer"}},
+            schema={
+                "$schema": "http://json-schema.org/draft-07/schema#",
+                "items": [{}],
+                "additionalItems": {"type": "integer"},
+            },
```
cuducos, line 50: `jsonschema` 4 now [supports newer drafts of JSON Schema](https://github.com/Julian/jsonschema/blob/main/CHANGELOG.rst#v400), so as I am **not** an expert in JSON Schemas, I opted to specify the previous draft (the one taken as default in `jsonschema` 3) instead of testing the new stuff of the newer drafts.

Many tests were failing due to the [situation described in this `jsonschema` issue](https://github.com/Julian/jsonschema/issues/852#issuecomment-933376870).

I think (a) testing new JSON Schema drafts is the role of `jsonschema`, not Spidermon's;
and (b) if we want to add tests specific to a given JSON Schema draft, this should be the focus of a new PR.

tox.ini
```diff
@@ -1,8 +1,11 @@
 [tox]
-envlist = py36,py37,py38,py39
+envlist = py{3.6,3.7,3.8,3.9}-jsonschema{3,4}
 skip_missing_interpreters = True
 
 [testenv]
+deps =
+    jsonschema3: jsonschema>=3.0.0,<4.0.0
+    jsonschema4: jsonschema>=4.0.0,<5.0.0
```
cuducos, line 8: This would only be needed if we want to support `jsonschema` versions 3 **and** 4 (and, in that case, I need to check how to run this in the CI yet).

PR #322, opened 2021-11-01
fix flaky test on test_validators_schematics.py::test_choices

lonly7star:
This PR aims to fix the flaky test on **test_validators_schematics.py::test_choices**  so the test could pass single run, multiple test run, and random test run. I'm doing this for a course project as a practice.

#### The result

The test won't pass on the single run with pytest, the following error raised
`   assert (False, {'a':...in choices']}) == (False, {'a':...[1, 2, 3].']})
At index 1 diff: {'a': ['Value not in choices'], 'b': ['Value not in choices']} != {'a': ["Value must be one of ['a', 'b']."], 'b': ['Value must be one of [1, 2, 3].']}`

#### Steps to reproduce the issue

1.  run the test file with `pytest test_validators_schematics.py`

#### Issue of the code

After examining the code I suspect the passed changes on `contrib::Validation::Validator` change how the error message is translated by `contrib::validation::schematics::translator.py`, which now raised a more detailed error message on `VALUE_NOT_IN_CHOICES` that caused an inconsistency on the testing error message.

#### Proposed solution
The convenience way is to changes the expect testing error message from `{'a': ['Value not in choices'], 'b': ['Value not in choices']}` to ` {'a': ["Value must be one of ['a', 'b']."], 'b': ['Value must be one of [1, 2, 3].']}`

### DIFF
--- a/tests/test_validators_schematics.py
+++ b/tests/test_validators_schematics.py
@@ -74,8 +74,8 @@ def test_choices():
         expected=(
             False,
             {
-                "a": [messages.VALUE_NOT_IN_CHOICES],
-                "b": [messages.VALUE_NOT_IN_CHOICES],
+                "a": ["Value must be one of ['a', 'b']."],
+                "b": ["Value must be one of [1, 2, 3]."],
             },
         ),
     )


VMRuiz:
Schematics is not longer supported

PR #323, opened 2021-11-29
Add `JobsComparisonMonitor`

### LINKED ISSUE #301: Built-in monitor to compare item count with previous jobs
This monitor is more useful when we are running the spider in [Scrapy Cloud](https://www.zyte.com/scrapy-cloud/). The objective is to compare the number of extracted items in current job with the number of items extracted in previous jobs.

- Get only `finished` jobs.
- Limit the number of jobs to be considered (configurable)
- if possible add more ways to decide which jobs to consider, not only the finished ones. This is useful when we have a `finished` job, but we want to exclude that execution for some reason (it failed y other reason for example)
- Configurable percentage to allow variations ("require that this job has at least 90% of items from the average 10 last jobs)

tcurvelo:
closes #301

It adds the `JobComparisonMonitor`, for checking for drops in scraped item count compared to previous jobs, when running jobs in [Scrapy Cloud](https://www.zyte.com/scrapy-cloud).
- allow filtering by jobs states (eg.: `finished`, `running`, `closespider_timeout`, etc)
- allow filtering by arbitrary tags 
- also updates the module that holds a singleton to the Scrapy Cloud API client.

----
TODO:
- [x] use scrapinghubclient instead
- [x] inherit from `BaseStatMonitor`
- [x] filter by tags


spidermon/contrib/scrapy/monitors.py
```diff
@@ -374,6 +380,53 @@ def test_check_if_field_coverage_rules_are_met(self):
         self.assertTrue(len(failures) == 0, msg=msg)
 
 
+@monitors.name("Jobs Comparison Monitor")
+class JobsComparisonMonitor(BaseScrapyMonitor):
```
rennerocha, line 384: **suggestion** You could use the new [BaseStatMonitor](https://spidermon.readthedocs.io/en/latest/monitors.html#base-stat-monitor) to create this monitor and define a custom `get_threshold` method that do all the logic of getting data from previous jobs.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -374,6 +380,53 @@ def test_check_if_field_coverage_rules_are_met(self):
...
+    """
+
+    @property
+    def client(self):
+        return hs
+
+    @monitors.name("Should not have a big drop in item count compared to previous jobs")
+    def test_if_item_count_from_previous_jobs_have_decreased(self):
+        # 1. get the number of jobs and check if monitor is enabled
+        number_of_jobs = self.crawler.settings.getint(SPIDERMON_JOBS_COMPARISON, 0)
```
rennerocha, line 405: **suggestion** To avoid the monitor to be executed without the proper settings, you could do the validation of the existence of the settings and raise a [NotConfigured](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/scrapy/monitors.py#L98)  exception when it is not. Doing that, we will receive a notification of the failure sooner.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -374,6 +380,53 @@ def test_check_if_field_coverage_rules_are_met(self):
...
+    You can configure which percentage of the previous item count is the minimum acceptable, by
+    using the setting ``SPIDERMON_JOBS_COMPARISON_THRESHOLD``. We expect a float number between
+    ``0.0`` and ``1.0``. The default is ``0.8``.
+
+    You can filter which jobs to compare based on their states using the
+    ``SPIDERMON_JOBS_COMPARISON_STATES`` setting. The default value is ``("finished",)``.
+    """
+
+    @property
+    def client(self):
```
rennerocha, line 399: **question** Where this property is being used?

spidermon/contrib/scrapy/monitors.py
```diff
@@ -374,6 +380,53 @@ def test_check_if_field_coverage_rules_are_met(self):
...
+
+    @monitors.name("Should not have a big drop in item count compared to previous jobs")
+    def test_if_item_count_from_previous_jobs_have_decreased(self):
+        # 1. get the number of jobs and check if monitor is enabled
+        number_of_jobs = self.crawler.settings.getint(SPIDERMON_JOBS_COMPARISON, 0)
+        if not number_of_jobs:
+            return
+
+        # 2. get the previous jobs meeting the defined states
+        states = self.crawler.settings.getlist(
```
rennerocha, line 410: **suggestion (non-blocking)** Instead of filtering just for finished status, perhaps we could have a list of tags we can use to filter.
For example, in a project, I have some jobs that can be `finished`, but because of some different reasons, I don't want to consider them as `valid`. So I set a `SUCCESS` tag on them. Give the choice to filter by tag may be interesting to avoid us to use bad data for comparison.

rennerocha (changes requested):
@tcurvelo thanks for the PR. I added some suggestions and questions.

spidermon/contrib/scrapy/monitors.py
tcurvelo: Sure thing. Will do that.

spidermon/contrib/scrapy/monitors.py
tcurvelo: Sure. This is a WIP yet. Filtering by tags was on the to-do list. Thanks, anyway.

spidermon/contrib/scrapy/monitors.py
tcurvelo: [A few lines below](https://github.com/scrapinghub/spidermon/pull/323/files#diff-9d830f5588694dea71ec34a158727c22a20d77630f7d8afd1705d966241ab00dR413).

It is just a wrapper for the hs client (which I'm still amending). I made it a property for an easier mocking.

spidermon/utils/zyte.py
```diff
@@ -0,0 +1,81 @@
+"""
+Module to hold a singleton to Zyte's Scrapy Cloud client.
```
rennerocha, line 2: **praise** This module was improved a lot with your changes.

**suggestion** However, these changes are good, but are not related entirely with the main purpose of your PR (the `JobsComparisonMonitor`). Is it possible to rebase these changes and create a separated PR just with the changes in the hubstorage module?

spidermon/contrib/scrapy/monitors.py
```diff
@@ -522,6 +530,78 @@ def test_execution_time(self):
...
+       `Zyte's Scrapy Cloud <https://www.zyte.com/scrapy-cloud/>`_.
+
+    Check for a drop in scraped item count compared to previous jobs.
+
+    You need to set the number of previous jobs to compare, using ``SPIDERMON_JOBS_COMPARISON``.
+    The default is ``0`` which disables the monitor. We use the average of the scraped items count.
+
+    You can configure which percentage of the previous item count is the minimum acceptable, by
+    using the setting ``SPIDERMON_JOBS_COMPARISON_THRESHOLD``. We expect a float number between
+    ``0.0`` and ``1.0``. The default is ``0.8``.
```
rennerocha, line 547: **suggestion** By default, our built-in monitors (see `ItemCountMonitor`, `ErrorCountMonitor`, etc) don't have a default value for the threshold, but they raise a `NotConfigured` if the setting is not provided. I'd prefer to keep it the same with this monitor, so we don't have a default value.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -522,6 +530,78 @@ def test_execution_time(self):
...
+
+        current_tags = json.loads(os.environ.get("SHUB_JOB_DATA", "{}")).get("tags")
+        if not current_tags:
+            return {}
+
+        tags_to_filter = set(desired_tags) & set(current_tags)
+        return sorted(tags_to_filter)
+
+    def get_threshold(self):
+        # 1. get the number of jobs and check if monitor is enabled
```
rennerocha, line 584: **suggestion** The method code is clear enough, so we don't need to add these comments.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -522,6 +530,78 @@ def test_execution_time(self):
...
+        number_of_jobs = self.crawler.settings.getint(SPIDERMON_JOBS_COMPARISON, 0)
+        if not number_of_jobs:
+            self.skipTest("Jobs comparison monitor is disabled")
+
+        # 2. get the previous jobs meeting the defined states and tags
+        states = self.crawler.settings.getlist(
+            SPIDERMON_JOBS_COMPARISON_STATES, ("finished",)
+        )
+        tags = self._get_tags_to_filter()
```
rennerocha, line 593: **suggestion** We don't need `tags` to be defined here. It could be something inside `_get_jobs`. This would make it easier to override `get_threshold` method if needed keeping the login to get job the same.

rennerocha (changes requested):
@tcurvelo added a few suggestions.

spidermon/utils/zyte.py
tcurvelo: Sure, no problem. https://github.com/scrapinghub/spidermon/pull/344

spidermon/contrib/scrapy/monitors.py
```diff
@@ -522,6 +530,78 @@ def test_execution_time(self):
...
+    that are also present in the current job.
+    """
+
+    stat_name = "item_scraped_count"
+    assert_type = ">="
+
+    def _get_jobs(self, states, number_of_jobs, tags=None):
+        return zyte.client.spider.jobs.list(
+            state=states,
+            count=number_of_jobs,
```
mushtaqak, line 563: I think this would only return a maximum of `1000` jobs because `jobs.list` uses `jobs.iter` internally and it only returns 1000 results by default
> [by default :meth:`Jobs.iter` returns maximum last 1000 results](https://github.com/scrapinghub/python-scrapinghub/blob/master/scrapinghub/client/jobs.py#L189)

cc: @rennerocha

curita:
@rennerocha hi! Bothering you again 🙈 This PR is currently stalled, but the related ticket is assigned for the next milestone so I imagine we'd like to move forward with it. Do you think it would be ok to assign someone from Zyte to work on it? If you'd like to continue with it instead no worries, we can work on something else, let us know what you'd like to do!

Issue #321, opened 2021-10-21
Base built-in monitor for stats validation


rennerocha:
**Problem**

A huge number of monitors follows the same pattern: we have one spider execution statistic and want to validate against a desired threshold. If we look the code of some [built-in monitors](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/scrapy/monitors.py) we can see that the code of them are almost the same (for example `ErrorCountMonitor`, `ItemCountMonitor`, `WarningCountMonitor` and some others).

Adding a new monitor is not too complicated, but we end with a lot of almost equal code that reduces the readability of the code and make it possible to introduce errors (because sometimes we do a lot of copy-paste to create our monitors).

**Solution**

Add a new base monitor class that allows us to create new monitors based on a value in the jobs statistics just defining a new class and a few class arguments. Doing that, we can reduce the amount of code we need to create to add a simple monitor like mentioned before.

**Draft of solution**

Creating a base stats monitor that allows us to create a new custom monitor with 5 lines of code:

```
class BaseStatsMonitor(Monitor):

    def test_validate_stats(self):
        # breakpoint()
        threshold = self.data.spider.settings.getint(self.threshold_setting, 10)
        value = self.data.stats.get(self.stats_key)
        error_msg = f"Invalid {self.stats_key} value. Expected: {threshold}. Obtained: {value}"
        assertions = {
            "GT": self.assertGreater,
            "GTE": self.assertGreaterEqual,
            "LT": self.assertLess,
            "LTE": self.assertLessEqual,
            "EQ": self.assertEqual,
            "NEQ": self.assertNotEqual,
        }
        assertion_method = assertions.get(self.assert_type)
        assertion_method(value, threshold, error_msg)

class ItemCountMonitor(BaseStatsMonitor):
    stats_key = "item_scraped_count"
    threshold_setting = "SPIDERMON_MIN_ITEMS"
    assert_type = "GTE"
    description = "Should extract the minimum amount of items"


class ValidationErrorsMonitor(BaseStatsMonitor):
    stats_key = "spidermon/validation/fields/errors
    threshold_setting = "SPIDERMON_MAX_VALIDATION_ERRORS
    assert_type = "LTE"
    description = "Limit the number of validation" 
```

rennerocha:
Closed by https://github.com/scrapinghub/spidermon/pull/325

PR #325, opened 2021-11-29
Create Base Stat Monitor class to help creating custom monitors

rennerocha:
Solves #321 

This PR includes a easier way to create custom monitors based on simple comparisons between a job stat and a threshold. We are talking usually about numerical values, but for EQUAL and NOTEQUAL, this can be strings as well.

A monitor that checks if a stat called `test_stat` is greater than or equal to a value defined in a project setting called `TEST_STAT_THRESHOLD`can be defined as follows:

```
    class TestStatMonitor(BaseStatMonitor):
        stat_name = "test_stat"
        threshold_setting = "TEST_STAT_THRESHOLD"
        assert_type = AssertionType.GTE
```

The threshold value can also be defined by a method, so we are able to create more complex rules to define what the threshold should be. Some ideas that can be accomplished using that method:
* Compare the number of items returned with the number of the latest job
* Define the threshold as a percentage of the number of input URLs against the number of requests
* etc

```
    class TestStatMonitor(BaseStatMonitor):
        stat_name = "test_stat"
        assert_type = AssertionType.GTE

        def get_threshold(self):
            # Do something to get the value
            return threshold
```

We still need to add the docs for this new feature and rewrite most of the built-in monitors to use this new structure.

### DIFF
--- a/spidermon/contrib/scrapy/monitors.py
+++ b/spidermon/contrib/scrapy/monitors.py
@@ -28,6 +28,54 @@ class BaseScrapyMonitor(Monitor, SpiderMonitorMixin):
 
 
 class BaseStatMonitor(BaseScrapyMonitor):
+    """Base Monitor class for stat-related monitors.
+
+    Create a monitor class inheriting from this class to have a custom
+    monitor that validates numerical stats from your job execution
+    against a configurable threshold.
+
+    As an example, we will create a new monitor that will check if the
+    value obtained in a job stat 'numerical_job_statistic' is greater than
+    or equal to the value configured in ``CUSTOM_STAT_THRESHOLD`` project
+    setting:
+
+    .. code-block:: python
+
+        class MyCustomStatMonitor(BaseStatMonitor):
+            stat_name = "numerical_job_statistic"
+            threshold_setting = "CUSTOM_STAT_THRESHOLD"
+            assert_type = ">="
+
+    For the ``assert_type`` property you can select one of the following:
+
+    ==  =====================
+    >   Greater than
+    >=  Greater than or equal
+    <   Less than
+    <=  Less than or equal
+    ==  Equal
+    !=  Not equal
+    ==  =====================
+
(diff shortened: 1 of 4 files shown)


spidermon/contrib/scrapy/monitors.py
```diff
@@ -27,6 +37,55 @@ def monitor_description(self):
...
+
+        return super().run(result)
+
+    def _get_threshold_value(self):
+        if hasattr(self, "get_threshold"):
+            return self.get_threshold()
+        return self.crawler.settings.get(self.threshold_setting)
+
+    def test_stat_monitor(self):
+        assertions = {
```
further-reading, line 69: I think we could simplify this by making a dictionary mapping the typical expression as a string to the assert method like so:

```python
def test_stat_monitor(self):
    assertions = {
           ">": self.assertGreater,
            ">=": self.assertGreaterEqual,
            "<": self.assertLess,
            "<=": self.assertLessEqual,
            "==": self.assertEqual,
            "!=": self.assertNotEqual
}
```

That way as a user I can do something like this for testing if the threshold if greater than or equal:

```python
class TestStatMonitor(BaseStatMonitor):
        stat_name = "test_stat"
        threshold_setting = "TEST_STAT_THRESHOLD"
        assert_type = ">="
```

Which might be clearer as we are using the standard expression for the comparison rather than going via a custom enum.

further-reading (approved):
LGTM overall. I have a suggestion on how to make it a little clearer from a user perspective but it is still fine as is!

spidermon/contrib/scrapy/monitors.py
rennerocha: @further-reading I liked your idea. It is much more easier for the user to know and define the type of the assertion. I will update as you suggested.

further-reading (commented):
LGTM :)

rennerocha:
Added the missing docs. This PR is now ready for real review and when approved, to be merged :-)

ogiaquino:
LGTM

PR #326, opened 2021-12-14
Use BaseStatMonitor to implement built-in item count monitor

rennerocha:
Use the new `BaseStatMonitor` to implement Item Count Monitor


further-reading (approved):
LGTM

PR #334, opened 2022-02-08
Use BaseStatMonitor to implement built-in downloader exception monitor

rennerocha:
All built-in monitors should use `BaseStatMonitor` when possible. So this PR will continue converting them. Also improving the documentation with a bit more details about the monitor.


Gallaecio (approved):

andersonberg (approved):
Nice job, good docstrings

PR #344, opened 2022-04-03
Rename module `hubstorage` to `zyte` and update it

tcurvelo:
These changes were originally part of #323.

### DIFF
--- a/spidermon/contrib/scrapy/extensions.py
+++ b/spidermon/contrib/scrapy/extensions.py
@@ -171,7 +171,7 @@ class Spidermon:
             runner.run(suite, **data)
 
     def _generate_data_for_spider(self, spider):
-        from spidermon.utils.hubstorage import hs
+        from spidermon.utils.zyte import client as hs
 
         return {
             "stats": self.crawler.stats.get_stats(spider),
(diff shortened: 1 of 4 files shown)


spidermon/contrib/scrapy/extensions.py
```diff
@@ -171,7 +171,7 @@ def _run_suites(self, spider, suites):
             runner.run(suite, **data)
 
     def _generate_data_for_spider(self, spider):
-        from spidermon.utils.hubstorage import hs
+        from spidermon.utils.zyte import client as hs
```
rennerocha, line 174: **question** Why define an alias for this import? We are not using it in other places, so `from spidermon.utils.zyte import client` and then `"job": client.job if client.available else None` looks fine to me.

rennerocha:
You got a drop in coverage probably because you kept `spidermon/utils/hubstorage.py`. As you were renaming the module you should use `git mv spidermon/utils/hubstorage.py spidermon/utils/zyte.py`, so we keep the history of the file. I believe that if you remove that now, git will understand that it was just a rename.

rennerocha (changes requested):

Issue #337, opened 2022-02-11
feature: Support for using SMTP as e-mail action


rennerocha:
There is already an old open PR for this action:
https://github.com/scrapinghub/spidermon/pull/204

Currently we only have the option to send e-mail reports using Amazon SES. If we want to use a different e-mail we have to create our own custom action. Considering that using `smtlib` to send email is easily configured in the majority of the email providers, a built-in action for that would be a good idea.

We also have a method in Scrapy to help us with that, that can be included:
https://docs.scrapy.org/en/latest/topics/email.html

We can use the old PR as reference or create a new one. It is up to the person that will develop this issue.

rennerocha:
Closed by https://github.com/scrapinghub/spidermon/pull/345

PR #345, opened 2022-04-06
Added SMTP email support

shafiq-muhammad:
Adding the support for SMTP email support

### DIFF
--- a/spidermon/contrib/actions/email/ses.py
+++ b/spidermon/contrib/actions/email/ses.py
@@ -58,7 +58,7 @@ class SendSESEmail(SendEmail):
                 recipients.extend(recipient_subset)
         return recipients
 
-    def send_message(self, message):
+    def send_message(self, message, **kwargs):
         client = boto3.client(
             service_name="ses",
             region_name=self.aws_region_name,
--- a/spidermon/contrib/actions/email/smtp.py
+++ b/spidermon/contrib/actions/email/smtp.py
@@ -46,7 +46,7 @@ class SendSmtpEmail(SendEmail):
         )
         return kwargs
 
-    def send_message(self, message):
+    def send_message(self, message, **kwargs):
         if not self.sender:
             raise NotConfigured("You must provide the sender.")
         if not self.to:
@@ -56,7 +56,11 @@ class SendSmtpEmail(SendEmail):
             return
 
         server = MailSender(self.smtp_host, self.sender, self.smtp_user, self.smtp_password,
-                            self.smtp_port, smtpssl=self.smtp_ssl)
+                            self.smtp_port, smtpssl=self.smtp_ssl, debug=bool(kwargs.get('debug')))
         server.send(
-            to=self.to, subject=message["Subject"], body=message.as_string(), cc=self.cc
+            to=self.to,
+            subject=message["Subject"],
+            body=message.as_string(),
+            cc=self.cc,
+            _callback=kwargs.get('_callback'),
         )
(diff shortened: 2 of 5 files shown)


curita:
Hi @shafiq-muhammad!

Some considerations:
1. The CI build for Python 3.9, those errors should be fixed
2. About adding tests, I found there's a flag `debug` that can be sent to `MailSender()` so no actual messages are sent ([reference](https://github.com/scrapy/scrapy/blob/ded28f7bb908090952c066a782c8518efaf2dd79/scrapy/mail.py#L35)). This is used in Scrapy's tests with a callback to catch the message ([reference](https://github.com/scrapy/scrapy/blob/master/tests/test_mail.py)). Maybe that's something we could replicate to add tests for the new `SendSmtpEmail()`?

    We could add a `debug` argument to `SendSmtpEmail.__init__()` or `SendSmtpEmail.send_message()`, and then pass that flag to `MailSender()`. We don't need to test `MailSender()` much as that's already handled by Scrapy, we can focus here on making sure we are instantiating it correctly and calling its send method as expected, and that the message gets "delivered", by using that callback or capturing the debug log messages.

Let me know if that helps and ping me for anything else I can assist you with!

spidermon/contrib/actions/email/smtp.py
```diff
@@ -0,0 +1,66 @@
...
+                "smtp_password": crawler.settings.get("SPIDERMON_SMTP_PASSWORD"),
+                "smtp_ssl": crawler.settings.get("SPIDERMON_SMTP_SSL")
+            }
+        )
+        return kwargs
+
+    def send_message(self, message, **kwargs):
+        if not self.sender:
+            raise NotConfigured("You must provide the sender.")
+        if not self.to:
```
rennerocha, line 52: **issue** Recipient check is [already validated](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/email/__init__.py#L57) inside `SendEmail` definition.

spidermon/contrib/actions/email/ses.py
```diff
@@ -54,7 +54,7 @@ def _get_recipients(self):
                 recipients.extend(recipient_subset)
         return recipients
 
-    def send_message(self, message):
+    def send_message(self, message, **kwargs):
```
rennerocha, line 57: **question** Why are you changing this class that is not related to the issue you are implementing?

rennerocha (changes requested):
Added some comments and suggestions into the PR.

Regarding the tests, I agree with @curita . We don't need to test the internals of `MailSender` that is already handled by Scrapy. You need to test if `MailSender.send()` is called using the right values according to the settings provided (search for assert_called_with in pytest docs).

Another test that is missing is to ensure that the `NotConfigured` is raised for each of the settings are not provided ([an example can be found here](https://github.com/scrapinghub/spidermon/blob/master/tests/contrib/actions/telegram/test_telegram_action.py#L50).

spidermon/contrib/actions/email/ses.py
shafiq-muhammad: I needed to add callback and setting some argument such as debug in MailSender while doing testing. So I've to change the base send_message and to make sure that this class doesn't break with that change. That's why I've changed this method.

spidermon/contrib/actions/email/smtp.py
shafiq-muhammad: Yes, That's correct. I'll remove check on that.

rennerocha:
This PR solves #337

rennerocha:
Hi @curita ! Yes, I am trying to finish this PR so we can have a new release of Spidermon (I need some new features!).
I added a few more settings options (that are available in Scrapy MailSender class) and added the documentation. It is only missing an update on the tests (current tests are testing MailSender, not the action). I should be finishing this by this week.

curita:
Thank you for the update and your work here again 🙇‍♀️ Sounds good then, if you need any help let us know!

PR #346, opened 2022-04-06
Allow Arbitrary kwargs when sending Slack messages using the slack Client postMessage method

### LINKED ISSUE #293: Allow pass extra optional arguments to Slack API
When sending a [Slack message](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/actions/slack/__init__.py#L111) we don't have an easy way to provide extra arguments to [postMessage](https://api.slack.com/methods/chat.postMessage) method (`mrkdwn` or `thread_ts` for example). It would be good to be able to provide extra arguments.

My use case is that I have an already created thread on  a monitoring Slack channel and I want that Spidermon notifications goes to a specific thread instead of going as a new message on the main channel.

ankitjavalkar:
Resolves #293

### DIFF
--- a/tests/contrib/actions/slack/test_slack_action.py
+++ b/tests/contrib/actions/slack/test_slack_action.py
@@ -1,5 +1,6 @@
 import sys
 import pytest
+from unittest.mock import MagicMock
 
 from spidermon.contrib.actions.slack import SlackMessageManager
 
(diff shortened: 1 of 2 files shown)


Gallaecio:
I think this is not enough. You are allowing arbitrary keyword arguments in a (private) method, but all methods that call it (including the public ones that are the ones users are expected to actually call) still do not allow those parameters.

ankitjavalkar:
@Gallaecio I have now added the **kwargs to function calls and public methods as well. I have also added tests using MagicMock.

tests/contrib/actions/slack/test_slack_action.py
```diff
@@ -58,3 +59,35 @@ def test_do_not_log_text_and_attach_when_fake_is_not_set(logger_info):
...
+    manager._client.chat_postMessage(
+        to=[], text='some_text', reply_broadcast=True,
+        thread_ts='some_thread_ts_value',
+        unfurl_links=False,
+    )
+    manager._client.chat_postMessage.assert_called_once_with(
+        to=[], text='some_text', reply_broadcast=True,
+        thread_ts='some_thread_ts_value',
+        unfurl_links=False,
+    )
```
Gallaecio, line 93: I think it would be best to create 1 test function per public method that now allows `**kwargs`, but in all of those test functions mock `manager._client.chat_postMessage`, then call the actual public method being tested (e.g. `manager.send_message`), and assert that the specific keyword-argument that is not part of the method signature reached the call to `manager._client.chat_postMessage`.

VMRuiz:
Closed in favor of https://github.com/scrapinghub/spidermon/pull/397/files

PR #352, opened 2022-08-02
Adding extra arguments for Slack Message like reply to thread & markup

muneer22:
Fixes https://github.com/scrapinghub/spidermon/issues/293

- Initial push as WIP.
- I have some more queries regarding this.

### DIFF
--- a/spidermon/contrib/actions/slack/__init__.py
+++ b/spidermon/contrib/actions/slack/__init__.py
@@ -14,8 +14,11 @@ logger = logging.getLogger(__name__)
 class SlackMessageManager:
     sender_token = None
     sender_name = None
+    thread_id = None
+    also_send_channel= None
+    mrkdwn = None
 
-    def __init__(self, sender_token=None, sender_name=None, fake=False):
+    def __init__(self, sender_token=None, sender_name=None, fake=False, thread_id=None, also_send_channel=None, mrkdwn=None):
         sender_token = sender_token or self.sender_token
         if not sender_token:
             raise NotConfigured(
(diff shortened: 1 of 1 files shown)


spidermon/contrib/actions/slack/__init__.py
```diff
@@ -110,14 +116,22 @@ def _send_user_message(
...
             attachments=self._parse_attachments(attachments),
             username=self.sender_name,
             icon_url=self._get_icon_url(),
+            thread_ts=self.thread_id,
+            # thread_ts = "1659182317.958909",
+            reply_broadcast=self.also_send_channel,
+            mrkdwn=self.mrkdwn
+        )
+
+        self.thread_id = (
```
curita, line 133: **issue:** if we reuse the `thread_id` all the Slack messages will be grouped under the same thread. This is a nice feature to have in certain projects, but it should probably be implemented separately and not as the default behavior, or it'd break backward compatibility.

rennerocha: The main reason of this issue was exactly to make it possible to send messages to the same thread (using `thread_ts` parameter). However, if we don't provide a thread_id, we shouldn't reuse the same for all messages.

curita: You are right, should have read the ticket description 😅 Makes sense to rely on the provided `thread_id`, and leave it to users to define it.

Gallaecio:
See also https://github.com/scrapinghub/spidermon/pull/346

VMRuiz:
Closed in favor of more generic solution  https://github.com/scrapinghub/spidermon/pull/397/files

Issue #353, opened 2022-08-02
Use itemadapter


kmike:
I've seen some code in projects, with a pipeline like this:

```py
# The pipeline needed to make Spidermon and attr items work nicely.
class ItemToDictTransformPipeline:
    def process_item(self, item, spider):
        adapter = ItemAdapter(item)
        return adapter.asdict()
```

This pipeline is not a proper solution, because it discards information about fields defined on the item, as well as field metadata. It means that exporting could work differently, e.g. CSV header may be set incorrectly.

I haven't checked why it's needed, but currently spidermon uses some scary code to convert items to dicts: https://github.com/scrapinghub/spidermon/blob/1838ec9e998c24fa64de38e50d5338ab4ce2b303/spidermon/contrib/scrapy/pipelines.py#L138 

It seems it can be simplified, sped up and made more robust by using https://github.com/scrapy/itemadapter.

Other methods also don't seem to support all the items which Scrapy supports, e.g. `_add_errors_to_item` only supports Item and dict. Again, this can be simplified and made more correct by using itemadapter.

I think it makes sense to check the whole contrib.scrapy, there are other issues like this, e.g. https://github.com/scrapinghub/spidermon/blob/1838ec9e998c24fa64de38e50d5338ab4ce2b303/spidermon/contrib/scrapy/extensions.py#L135

Gallaecio:
Makes sense after https://github.com/scrapinghub/spidermon/pull/358

PR #358, opened 2022-08-12
Adding use of ItemAdapter to prevent assumptions of item nature

further-reading:
https://github.com/scrapinghub/spidermon/issues/353


kmike:
hey! Would it be possible to add tests for it?

further-reading:
It's still a work in progress haha, of course there'll be tests. We can hold off on the reviews until its ready.

spidermon/contrib/scrapy/pipelines.py
```diff
@@ -136,25 +133,16 @@ def find_validators(self, item):
...
-            if self.errors_field not in item.__class__.fields:
-                item.__class__.fields[self.errors_field] = Field()
-            if self.errors_field not in item._values:
-                item[self.errors_field] = defaultdict(list)
-        except AttributeError:
-            # The item is just a dict object instead of a Scrapy.Item object
-            if self.errors_field not in item:
-                item[self.errors_field] = defaultdict(list)
+        data = ItemAdapter(item)
+        if self.errors_field not in data.keys():
```
Gallaecio, line 141: :lipstick: I wonder if `in data:` would work as well.

Gallaecio (approved):
Looks good to me.

I would however add `itemadapter` as a dependency now. Things will work either way because of the dependency with Scrapy, but it makes things more future-proof (e.g. in the unlikely case that Scrapy drops itemadapter as a dependency).

wRAR (approved):

spidermon/contrib/scrapy/extensions.py
```diff
@@ -137,7 +138,8 @@ def _count_item(self, item, skip_none_values, item_count_stat=None):
             item_count_stat = f"spidermon_item_scraped_count/{item_type}"
             self.crawler.stats.inc_value(item_count_stat)
 
-        for field_name, value in item.items():
+        data = ItemAdapter(item)
+        for field_name, value in data.items():
```
elacuesta, line 142:
```suggestion
        for field_name, value in ItemAdapter(item).items():
```
Nitpick 1

spidermon/contrib/scrapy/pipelines.py
```diff
@@ -136,25 +133,16 @@ def find_validators(self, item):
...
 
     def _convert_item_to_dict(self, item):
-        serialized_json = BytesIO()
-        exporter = JsonLinesItemExporter(serialized_json)
-        exporter.export_item(item)
-        data = json.loads(to_unicode(serialized_json.getvalue(), exporter.encoding))
-        serialized_json.close()
-        return data
+        data = ItemAdapter(item)
+        return data.asdict()
```
elacuesta, line 137:
```suggestion
        return ItemAdapter(item).asdict()
```
Nitpick 2

spidermon/contrib/scrapy/pipelines.py
elacuesta: It would because `ItemAdapter` implements `MutableMapping`. It would also be a tiny bit more efficient if I'm not mistaken.

setup.py
```diff
@@ -12,6 +12,7 @@
     "scrapy",
     "slack-sdk",
     "twisted>=19.7.0",
+    "itemadapter==0.7.0",
```
Gallaecio, line 15: In packages meant to be used as dependencies by others the dependency strategy is to allow as many versions of dependencies as possible, as to not lock users into specific versions, which can make it extremely hard to combine different packages with a shared dependency.

So here we should either go for no version at all:

 ```suggestion
    "itemadapter",
```

Or find out what minimum version of itemadapter Spidermon needs, though given the features used I think even the first version of itemadapter may work.

Gallaecio:
@further-reading Unless you plan on further changes, i.e. https://github.com/scrapinghub/spidermon/pull/358#discussion_r968693785, we can merge as far as I am concerned.

PR #359, opened 2022-08-18
Add new feature to limit number of records in stats for SPIDERMON_ADD_FIELD_COVERAGE setting.

PyExplorer:
In some cases, we have a huge number of fields in one item (>500). Enabling SPIDERMON_ADD_FIELD_COVERAGE adds a record for every field in stats and this raises the issue like "Value exceeds max encoded size of 65536 bytes" in Scrapy Cloud dashboard.

The idea is to limit the number of records in the stats to only fields that we actually want to monitor in SPIDERMON_FIELD_COVERAGE_RULES.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -142,7 +142,7 @@ field following this format:
       'spidermon_field_coverage/dict/field_1/nested_field_1_2': 0.5,
       'spidermon_item_scraped_count/dict/field_2': 0.5,
 
-SPIDERMON_LIMIT_FIELD_COVERAGE_BY_FIELD_COVERAGE_RULES
+SPIDERMON_FIELD_COVERAGE_SKIP_WITHOUT_FIELD_COVERAGE_RULES
 ----------------------------
 Default: ``False``
 
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -143,7 +143,7 @@ field following this format:
       'spidermon_item_scraped_count/dict/field_2': 0.5,
 
 SPIDERMON_FIELD_COVERAGE_SKIP_WITHOUT_FIELD_COVERAGE_RULES
-----------------------------
+----------------------------------------------------------
 Default: ``False``
 
 When enabled, Spidermon will add statistics about the number of items scraped and coverage only for fields that
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -142,6 +142,76 @@ field following this format:
       'spidermon_field_coverage/dict/field_1/nested_field_1_2': 0.5,
       'spidermon_item_scraped_count/dict/field_2': 0.5,
 
+SPIDERMON_LIMIT_FIELD_COVERAGE_BY_FIELD_COVERAGE_RULES
+----------------------------
+Default: ``False``
+
+When enabled, Spidermon will add statistics about the number of items scraped and coverage only for fields that
+exist in setting SPIDERMON_FIELD_COVERAGE_RULES in following format:
+
+``'spidermon_item_scraped_count/<item_type>/<field_name>': <item_count>``
+``'spidermon_field_coverage/<item_type>/<field_name>': <coverage>``
+
+Using this setting will help to limit number of records in statistic in case of huge number of fields (>500)
+in the item.
+
+.. note::
+
+   Nested fields are also supported. For example, if your spider returns these items:
+
+   .. code-block:: python
+
+      [
+        {
+          "field_1": {
+            "nested_field_1_1": "value",
+            "nested_field_1_2": "value",
+          },
+        },
+        {
+          "field_1": {
+            "nested_field_1_1": "value",
+          },
+          "field_2": "value"
+        },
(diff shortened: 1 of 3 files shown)


VMRuiz:
I have mixing feelings here. On one hand, I understand this may be useful feature to reduce overall noise in stats. 
On the other hand, it looks like we are adding a pretty specific feature just to overcome a limitation on Scrapy Cloud. An alternative solution would be to override the stat export for Scrapy Cloud to ignore the non-desired fields from the spider.

I would like to heard more opinions about this. cc. @Gallaecio @kmike @rennerocha @wRAR

rennerocha:
Adding the stats just of the fields that have a field coverage rule is not ideal. It won't prevent you to have the Scrapy Cloud limitation problem if your spider adds more stats unrelated to field coverage (I had a project like that when I worked at Zyte). It will be just a **palliative** solution as we can't guarantee that we will not have more than 65536 bytes of stats just removing the field coverage stats.

I know that this is annoying, but this PR is not a solution for a Scrapy or Spidermon limitation, but a solution for a third-party application limitation (as we may be running Scrapy spiders anywhere without that limitation). So, in my opinion, I agree with Victor, that this is a too specific feature just to overcome a problem in Scrapy Cloud.

Some alternatives solution I can see:

* **My preference** - Every time we start a job in Scrapy Cloud, the default stats collector (`STATS_CLASS`) is overrided to [sh_scrapy.stats.HubStorageStatsCollector](https://github.com/scrapinghub/scrapinghub-entrypoint-scrapy/blob/master/sh_scrapy/stats.py#L18). You could modify this stats collector to not persist the field coverage related stats to SC. That way the feature will be limited to a project that is 100% specific to people running the Spiders in Scrapy Cloud.

* Remove the dependency of `SPIDERMON_ADD_FIELD_COVERAGE` setting from `FieldCoverageMonitor` monitor. This would require that we change how we [keep the items returned](https://github.com/scrapinghub/spidermon/blob/3ef06ebf5c6d97ae45b199bc6acfbd322edad23b/spidermon/contrib/scrapy/extensions.py#L134), what we [store in the stats](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/scrapy/extensions.py#L151) and the [data used](https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/scrapy/monitors.py#L493) by the monitor.

rennerocha:
@PyExplorer just a nitpick, it would make easier for me to review code if you create the PR from a fork in your personal account (not from the main repository) :-)

Gallaecio:
If we forget about Scrapy Cloud altogether, I think it may be a valid request to want control over which fields Spidermon tracks. From that point of view, I am not against going this route.

However, for the stated goal, I agree with @rennerocha‘s preferred solution, modifying the special `STATS_CLASS` that Scrapy Cloud uses to communicate stats to Scrapy Cloud may be the best path. We could extend the upstream class to support e.g. a list of regular expressions to exclude stats, for example; it should not be too difficult to implement, and it would be flexible enough to also support other instances of the issue unrelated to Spidermon.

docs/source/settings.rst
```diff
@@ -142,6 +142,76 @@ field following this format:
       'spidermon_field_coverage/dict/field_1/nested_field_1_2': 0.5,
       'spidermon_item_scraped_count/dict/field_2': 0.5,
 
+SPIDERMON_FIELD_COVERAGE_SKIP_WITHOUT_FIELD_COVERAGE_RULES
```
Gallaecio, line 145: If we proceed, I think we need a shorter setting name.

I wonder if it would not be too messy to support a special value for `SPIDERMON_ADD_FIELD_COVERAGE`, e.g. `"rules"`.

Otherwise, what about `SPIDERMON_FIELD_COVERAGE_RULES_ONLY`, or something else similarly short?

spidermon/contrib/scrapy/extensions.py
```diff
@@ -157,8 +174,20 @@ def item_scraped(self, item, response, spider):
...
+        )
         self.crawler.stats.inc_value("spidermon_item_scraped_count")
-        self._count_item(item, skip_none_values)
+
+        if skip_fields_without_rules and not field_coverage_rules:
+            return
+
+        self._count_item(
+            item, skip_none_values, field_coverage_rules, skip_fields_without_rules
+        )
```
Gallaecio, line 190: We could make this cleaner by first defining `skip_fields` as `None` if `skip_fields_without_rules` is `None` or as the set of fields to ignore otherwise, hence only passing 1 variable forward.

rennerocha:
Just removing these stats for this specific monitor will not solve the problem. We don't know if other monitors, custom stats or logs added by the spider developers will not surpass the Scrapy Cloud limit. Also, this is not a problem with Spidermon's use of the stats collection feature of Scrapy, but a limitation of a third-party system.

Considering that there is a solution that will require a change in Scrapy Cloud stats collector only, and as @Gallaecio mentioned, would open room for more fine-grained control there, I will close this PR.

PR #364, opened 2022-10-04
Fix compatibility issues with jsonschema>=4

aaneto:
I have managed to pass the tests after updating jsonschema. The only issue I have is that Python 3.6 is having dependency issues with jsonschema==4.

I'll try to make the changes compatible with both jsonschema4 and 3, so that Python3.6 is still compatible.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install tox codecov
       - name: Run tests
-        run: tox -e ${{ matrix.tox || "py" }}
+        run: tox -e ${{ matrix.tox || 'py' }}
       - name: Run docs and formatting tests
         if: ${{ matrix.python-version == 3.10 }}
         run: |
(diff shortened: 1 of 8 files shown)


wRAR (approved):

aaneto:
I don't believe I see anything else that needs doing, I'll remove the WIP tag so that others can give feedback more freely.

Gallaecio (approved):
Looks good to me.

I do wonder if we should implement a test job that uses the minimum supported versions of stuff, like we do in Scrapy, to make sure we do not break backward compatibility with older jsonschema as we add support for newer ones.

aaneto:
Great :)

If I'm not mistaken all that we need now is for someone with merge permission to accept this PR, correct?

Gallaecio:
> I do wonder if we should implement a test job that uses the minimum supported versions of stuff, like we do in Scrapy, to make sure we do not break backward compatibility with older jsonschema as we add support for newer ones.

Let me implement this, then we can merge.

Gallaecio:
Thanks!

aaneto:
Oh, sorry! That's what you meant by testing job, I thought it was a job on a cloud or something like that.

In any case, thanks everyone! :)

PR #372, opened 2022-12-09
Allow to skip execution of FieldCoverageMonitor if no item is returned

### LINKED ISSUE #366: Allow to skip execution of FieldCoverageMonitor if no item is returned
This is a fork of issue #339 .

`FieldCoverageMonitor` requires that we have at least one item returned, or else it will fail. However there are situations when we accept that a spider return no items, but if items are returned, we want to ensure our field coverage. 

In a project that are in production, we have dozens of real estate websites searching for locations in different neighborhoods daily. Some neighborhoods sometimes doesn't return any item in a specific day, but when data is returned we want to ensure that all the fields has the right coverage.

One possible solution (that looks simpler to implement) would be add a new setting called `SPIDERMON_FIELD_COVERAGE_SKIP_IF_NO_ITEM` (or something similar). If set to True and no item is returned, field coverage will not be checked, otherwise, it will work as it is now.

VladAlex98:
Closes #366 

Added another setting called `SPIDERMON_FIELD_COVERAGE_SKIP_IF_NO_ITEM` for `FieldCoverageMonitor`:
- If this setting is enabled and we don't have any items scraped, it will not raise error anymore, it will send a message instead.

### DIFF
--- a/spidermon/contrib/scrapy/monitors/monitors.py
+++ b/spidermon/contrib/scrapy/monitors/monitors.py
@@ -424,7 +424,7 @@ class FieldCoverageMonitor(BaseScrapyMonitor):
         return super().run(result)
 
     def test_check_if_field_coverage_rules_are_met(self):
-        skip_no_items = self.crawler.settings.get(
+        skip_no_items = self.crawler.settings.getbool(
             "SPIDERMON_FIELD_COVERAGE_SKIP_IF_NO_ITEM", False
         )
         items_scraped = self.data.stats.get("item_scraped_count", 0)
--- a/spidermon/contrib/scrapy/monitors/monitors.py
+++ b/spidermon/contrib/scrapy/monitors/monitors.py
@@ -424,8 +424,10 @@ class FieldCoverageMonitor(BaseScrapyMonitor):
         return super().run(result)
 
     def test_check_if_field_coverage_rules_are_met(self):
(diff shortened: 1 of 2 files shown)


rennerocha (dismissed):
Hello @VladAlex98 ,
Thanks for your contribution. I added a suggestion so the logs returned for the user would looks nicer and more consistent with other monitor results logs that we have in other built-in monitors.
Also, do not forget to add unit tests for your change and update the documentation of FieldCoverageMonitor adding this new setting.

VladAlex98:
Hi @rennerocha, thanks for the code review. I updated the code, added docs and some tests. Please let me know if you need to add something else.

VMRuiz:
Hello @rennerocha , is this good now?

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -417,9 +420,17 @@ def run(self, result):
             raise NotConfigured(
                 "To enable field coverage monitor, set SPIDERMON_ADD_FIELD_COVERAGE=True in your project settings"
             )
+
         return super().run(result)
 
     def test_check_if_field_coverage_rules_are_met(self):
+        skip_no_items = self.crawler.settings.get(
```
curita, line 427: **suggestion:** We should use `.getbool()` here instead, so we can parse boolean-like strings as valid values too. That way, we'll support setting this value in ScrapyCloud or the command line.

curita (approved):
Changes look good!

I mentioned a small suggestion, but otherwise, the functionality makes sense to me.

PR #373, opened 2022-12-15
Change Ubuntu version on workflow settings.

mrwbarg:
Our pipeline hasn't been running for Python 3.6. This is due to an issue with github actions (https://github.com/actions/setup-python/issues/544). Fixing the ubuntu version to 20.04 should fix the problem according to the comments.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


curita:
@Gallaecio @rennerocha could you check these changes when you have the time? The CI checks don't pass for that version of python otherwise and affect all recent PRs.

curita (approved):

PR #378, opened 2023-01-12
Update `BaseStatMonitor`'s `test_stat_monitor` to work with Scrapy Cloud

### LINKED ISSUE #374: Error when threshold values are set via ScrapyCloud settings
The following error occurs when the spider is running on [ScrapyCloud](https://zyte.com/scrapy-cloud/) and the monitor threshold value is set via ScrapyCloud settings. The threshold retrieved from ScrapyCloud settings is a string, while the value retrieved from the job stats is an integer, therefore the comparisson fails.

```
======================================================================
ERROR: Extracted Items Monitor/test_stat_monitor
  File "/app/python/lib/python3.10/site-packages/spidermon/contrib/scrapy/monitors.py", line 144, in test_stat_monitor
    assertion_method(
  File "/usr/local/lib/python3.10/unittest/case.py", line 1248, in assertGreaterEqual
    if not a >= b:
TypeError: '>=' not supported between instances of 'int' and 'str'
```

If the monitor threshold is set only in the code (and not in ScrapyCloud), the value is retrieved as integer/float and no error is threw.

It seems it started when [this PR](https://github.com/scrapinghub/spidermon/pull/325) was merged.

The method used to [get the threshold value](https://github.com/scrapinghub/spidermon/pull/325/files#diff-9d830f5588694dea71ec34a158727c22a20d77630f7d8afd1705d966241ab00dR123) is `get()`, while it was using `getint()` before. It affects only monitors that are inheriting from the `BaseStatMonitor` class and don't implement a custom `get_threshold()` method.

mrwbarg:
Fixes #374.

When reading threshold values from the settings, the `BaseStatMonitor` [used ](https://github.com/scrapinghub/spidermon/pull/325/files#diff-9d830f5588694dea71ec34a158727c22a20d77630f7d8afd1705d966241ab00dR123) the `settings.get()` function. This works fine if the value is set in the project's settings as the data type is preserved when reading it. However, when getting the settings from a Scrapy Cloud project, using the bare `get()` function causes the value to be read as a string, which break comparisions.

The monitor was updated to use by default the `getfloat` function. A variable was added which allows setting the threshold data type (either `int` or `float`) along with a property that picks the correct _get_ function depending on the data type.

Two new unit tests were added to test for the cases where the values are actual strings.

Despite there being other monitors that inherit from BaseStatMonitor, no further changes were necessary as they'll benefit from the default `getfloat()` function.

### DIFF
--- a/spidermon/contrib/scrapy/monitors.py
+++ b/spidermon/contrib/scrapy/monitors.py
@@ -38,7 +38,10 @@ class BaseStatMonitor(BaseScrapyMonitor):
 
     Create a monitor class inheriting from this class to have a custom
     monitor that validates numerical stats from your job execution
-    against a configurable threshold.
+    against a configurable threshold. If this threshold is set in Scrapy
+    Cloud (and not it the spider settings), the setting is read as a
+    string and converted to ``threshold_datatype`` type (default is
+    float).
 
     As an example, we will create a new monitor that will check if the
     value obtained in a job stat 'numerical_job_statistic' is greater than
@@ -100,6 +103,17 @@ class BaseStatMonitor(BaseScrapyMonitor):
     """
 
     fail_if_stat_missing = True
+    threshold_datatype = float
+
+    @property
+    def _get_threshold_setting(self):
+
+        datatype_to_function = {
+            int: self.crawler.settings.getint,
+            float: self.crawler.settings.getfloat,
+        }
+
+        return datatype_to_function[self.threshold_datatype]
 
     def run(self, result):
         has_threshold_config = any(
@@ -126,7 +140,7 @@ class BaseStatMonitor(BaseScrapyMonitor):
     def _get_threshold_value(self):
         if hasattr(self, "get_threshold"):
             return self.get_threshold()
-        return self.crawler.settings.get(self.threshold_setting)
+        return self._get_threshold_setting(self.threshold_setting)
 
     def test_stat_monitor(self):
         assertions = {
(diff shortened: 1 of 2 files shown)


spidermon/contrib/scrapy/monitors.py
```diff
@@ -38,7 +38,10 @@ class BaseStatMonitor(BaseScrapyMonitor):
 
     Create a monitor class inheriting from this class to have a custom
     monitor that validates numerical stats from your job execution
-    against a configurable threshold.
+    against a configurable threshold. If this threshold is set in Scrapy
```
VMRuiz, line 41: I would replace mention to Scrpay Cloud to something else more general like passing settings from the CLI

spidermon/contrib/scrapy/monitors.py
```diff
@@ -100,6 +103,17 @@ class MyCustomStatMonitor(BaseStatMonitor):
     """
 
     fail_if_stat_missing = True
+    threshold_datatype = float
```
VMRuiz, line 106: Is there any risk on backward compatibility with all the `BaseStatMonitor` already implemented? 
Both in this package and customs that developers may have created.

For example, If I have already defined a threshold vallue as `int`, this feature would make spidermon to read the value as `float` instead. I'm not sure if that may have unexpected side effects

mrwbarg: I feel like reading an int as a float (in the monitors` use cases) will never cause an issue whilst the opposite wouldn't be true. 
For example: if a threshold is 135 and is read as 135.0, I reckon that wouldn't make a functional difference. If that's the case, what other option do we have? I think reading a float by default is the most backwards compatible thing we can do.

VMRuiz: My concern is that when testing for equality, there could be some unexpected round-off errors that caused that for some value `n`, `int(n) != float(x)`. 

However, I couldn't find any examples where this may happens.

VMRuiz (approved):

rennerocha (approved):

PR #380, opened 2023-01-17
Add Percentage Thresholds to `UnwantedHTTPCodesMonitor`

### LINKED ISSUE #376: idea: Percentage thresholds for UnwantedHTTPCodesMonitor
## Background

At Zyte, we have internal custom monitors that check the percentage of response failures and the percentage of ignored responses (`[private]` [reference](https://github.com/zytedata/spidermon-monitors-data-ops)).

Setting percentage thresholds instead of fixed max counts has been beneficial in projects where the number of initial requests varies. @gutsytechster suggested we could also use percentage thresholds for the unwanted HTTP status codes (`[private]` [thread](https://zytegroup.slack.com/archives/C0217MC5E4B/p1671181713529759)).

## Current Status

At the moment, UnwantedHTTPCodesMonitor thresholds can be either set via a general `SPIDERMON_UNWANTED_HTTP_CODES_MAX_COUNT` or a `SPIDERMON_UNWANTED_HTTP_CODES` dictionary, where the keys are the HTTP codes and the values are the max allowed responses with those status codes.

## Suggestion

We could add an alternative `SPIDERMON_UNWANTED_HTTP_CODES_THRESHOLDS` setting (or continue overloading `SPIDERMON_UNWANTED_HTTP_CODES`, but that might be confusing) for when we want to set percentage thresholds for unwanted response codes instead of fixed max counts. The option to set max counts should still exist.

mrwbarg:
Closes #376 

Added option to set thresholds as a percentage of total spider requests' for the `UnwantedHTTPCodesMonitor`.
The implementation follows the discussions on the issue and alos @curita's suggestions.

Added relevant unit tests.


VMRuiz (approved):
Looks good to me. I would like the percentage support to be expanded to others monitors, but we can do it in a new PR.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -288,6 +288,24 @@ class UnwantedHTTPCodesMonitor(BaseScrapyMonitor):
             500: 0,
         }
 
+    Furthermore, instead of being a numeric value, the code accepts a dictionary which can
+    contain any of two keys: ``max_count`` and ``percentage``. The former refers to an
+    absolute value and works the same way as setting an integer value. The latter refers
+    to a percentage of the total number of requests the spider made. If both are set, the
+    monitor will pick the smallest of the two values. If none are set, it will default to
```
Gallaecio, line 295: > If both are set, the monitor will pick the smallest of the two values.

I think this could lead to confusion. e.g. to think that, in the case of `{"max_count": 100, "percentage": 0.5}`, `max_count` is always ignored because `0.5 < 100`.

What about something in the lines of:

> If both are set, the monitor will fail if any of the conditions are met.

Which is what the implementation achieves, even if it is through the use of `min(max_count or requests, percentage*requests)`.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -288,6 +288,24 @@ class UnwantedHTTPCodesMonitor(BaseScrapyMonitor):
             500: 0,
         }
 
+    Furthermore, instead of being a numeric value, the code accepts a dictionary which can
+    contain any of two keys: ``max_count`` and ``percentage``. The former refers to an
```
Gallaecio, line 292: For consistency, shouldn’t it be `max_percentage`?

spidermon/contrib/scrapy/monitors.py
```diff
@@ -311,12 +329,31 @@ def test_check_unwanted_http_codes(self):
...
+                            percentual_max_errors * requests
+                            if percentual_max_errors
+                            else requests
+                        ),
+                    )
+
             msg = (
                 "Found {} Responses with status code={} - "
-                "This exceed the limit of {}".format(count, code, max_errors)
+                "This exceeds the limit of {}".format(count, code, max_errors)
```
Gallaecio, line 356: I think in the case of a percentage we might want to indicate the source percentage in addition to that that translates into, e.g. `limit of {} ({}% of {} total requests)`.

spidermon/contrib/scrapy/monitors.py
mrwbarg: Makes sense! Will update the text.

spidermon/contrib/scrapy/monitors.py
```diff
@@ -311,12 +329,45 @@ def test_check_unwanted_http_codes(self):
...
+
+                    max_errors = min(
+                        absolute_max_errors if absolute_max_errors else requests,
+                        calculated_percentage_errors,
+                    )
+
+                    percentage_trigger = max_errors == calculated_percentage_errors
+
+            stat_message = (
+                "This exceeds the limit of {} ({}% of {} total requests)".format(
```
Gallaecio, line 361: `{}%` will render as 0.5% instead of 50% for 0.5. See e.g. https://stackoverflow.com/a/23764798/939364

Also, assuming I did not get it wrong, maybe you can update tests so that they would have caught this issue.

mrwbarg: Hmm. I don't think it will because I added a `* 100` after it. I updated a couple cases just to be safe!

Gallaecio (approved):
Great!

VMRuiz (commented):
It took me a bit to understand why each calculation is needed inside the monitor, maybe some comments explaining why would help

mrwbarg:
> It took me a bit to understand why each calculation is needed inside the monitor, maybe some comments explaining why would help

Done

VMRuiz (approved):

PR #383, opened 2023-01-24
Move Zyte related code to own module

### LINKED ISSUE #382: Separate Zyte related code from core
Considering that Spidermon was created at Zyte (former Scrapinghub), it is understandable that we have monitors and actions that are related to services provided by it. However, considering that Spidermon can be used by users that are not running their spiders at Zyte infrastructure, it would be a good refactoring to keep every code that is Zyte-related only inside a separated module.

My first suggestion is to move everything (actions, monitors, mixins and utils) that are spread in the project into a separated `spidermon.contrib.zyte.*` module, so it will be clear to everyone that anything imported from this module is related to Zyte and may require that the extension is being executed inside Zyte infrastructure.

I found Zyte-only related code in the following files:
- spidermon/spidermon/utils/zyte.py
- spidermon/contrib/stats/statscollectors/sc_collections.py
- spidermon/contrib/scrapy/monitors.py
- spidermon/contrib/scrapy/monitors/mixins/job.py
- spidermon/contrib/actions/jobs/tags.py

Tasks to accomplish 
- [ ] Create `spidermon.contrib.zyte` module
- [ ] Move all related code to this module
- [ ] Avoid backward incompatibility adding a deprecated warning if we try to import in old locations (but allows it)
- [ ] Update documentation

rennerocha:
This fix #382

### DIFF
--- a/spidermon/contrib/actions/jobs/tags.py
+++ b/spidermon/contrib/actions/jobs/tags.py
@@ -1,48 +1,5 @@
-from spidermon.core.actions import Action
-from spidermon.exceptions import NotConfigured
-
-
-class JobTagsAction(Action):
-    tags = []
-    tag_settings = None
-
-    def __init__(self, tags=None):
-        super().__init__()
-        tags = tags or self.tags
-        self.tags = tags if isinstance(tags, list) else [tags]
-
-    @classmethod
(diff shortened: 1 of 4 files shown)


Gallaecio:
What about logging a deprecation warning when the old modules or classes are used?

PR #384, opened 2023-01-26
Addition of SPIDERMON_MONITOR_SKIPPING_RULES 

shafiq-muhammad:
- added SPIDERMON_MONITOR_SKIPPING_RULES so that we can skip any monitor with some condition based on stats value

Example of adding rule in the spider settings using stats value condition
```
"SPIDERMON_MONITOR_SKIPPING_RULES": {
            "Field Coverage Monitor": [["item_scraped_count", "==", 0]],
        }
```

Example of adding rule in the spider settings using custom function
```
"SPIDERMON_MONITOR_SKIPPING_RULES": {
            "Field Coverage Monitor": [skip_function],
        }

def skip_function(monitor):
    return "item_scraped_count" not in monitor.data.stats
```


curita:
Hi @shafiq-muhammad! Wonder if it might be better to discuss the design in the ticket itself 🤔 We could agree on how this feature would look first and involve Spidermon maintainers, and then proceed with the implementation.

I visualize this feature in these two ways:
1. For `skip_if` to be a decorator, which we can add over the monitors
    * This is a design preference, but I thought it would be nice for it to mimic the `skip` decorator from pytest.
2. Rather than defining static rules over the stats, `skip_if` could ask for a function. The function could take the monitor class as a parameter (we have access to the stats, spider, etc., inside it) and return True or False. If it returns True, that monitor will be skipped, and when it returns False, it will run as usual.

But that's just what I'd like to see on my end. We should gather some more feedback to help us define this.

curita:
I kept thinking about it, and now the external `SPIDERMON_MONITOR_SKIPPING_RULES` setting makes sense to me 🙏

The only thing I'd add is the support for user-defined functions for rules, but otherwise, I think we could go this route to add skip rules for any monitor we'd like without an explicit `@skip_if` decorator.

I left a comment in the original ticket for reference.

spidermon/contrib/scrapy/monitors/base.py
```diff
@@ -1,18 +1,86 @@
...
+    ops = {"==": operator.eq, "<": operator.lt, ">": operator.gt}
 
     @property
     def monitor_description(self):
         if self.__class__.__doc__:
             return self.__class__.__doc__.split("\n")[0]
         return super().monitor_description
 
+    """ 
+        The montior inherited by BaseScrapyMonitor can be skipped based on
```
Gallaecio, line 23: This docstrings seems misplaced.

spidermon/contrib/scrapy/monitors/base.py
```diff
@@ -1,18 +1,86 @@
...
 from spidermon.exceptions import NotConfigured
 
 from ...monitors.mixins.spider import SpiderMonitorMixin
 
+logger = logging.getLogger(__name__)
+
 
 class BaseScrapyMonitor(Monitor, SpiderMonitorMixin):
     longMessage = False
+    ops = {"==": operator.eq, "<": operator.lt, ">": operator.gt}
```
Gallaecio, line 14: Maybe we should support the same operations as https://github.com/scrapinghub/spidermon/blob/e08588691f31e13f2e10233f2711632ad2d49294/spidermon/contrib/scrapy/monitors/base.py#L194-L201, and maybe we could add a test that verifies that they keywords in these 2 dicts remain the same (i.e. so that, if we add an operation to one dict, we are reminded to also add it to the other).

VMRuiz (changes requested):
Could you please add tests for new lines?

VMRuiz (approved):

VMRuiz:
@Gallaecio  are we good to merge?

spidermon/contrib/scrapy/monitors/base.py
```diff
@@ -1,18 +1,93 @@
...
 
+    def run(self, result):
+        if self.check_if_skip_rule_met():
+            logger.info(f"Skipping {self.name} monitor")
+            return
+
+        return super().run(result)
+
+    """ 
+    The montior inherited by BaseScrapyMonitor can be skipped based on
```
Gallaecio, line 37: “The montior inherited by BaseScrapyMonitor” is `Monitor`. Did you mean the following?

```suggestion
    A monitor that inherits BaseScrapyMonitor can be skipped based on
```

spidermon/contrib/scrapy/monitors/base.py
```diff
@@ -1,18 +1,93 @@
...
+    conditions given in the settings. The purpose is to skip monitor 
+    based on stats value or any custom function. A scenario could be skipping
+    the Field Coverage Monitor when spider produced no items. Following is the
+    code block of examples of how we can configure the skip rules in settings.
```
Gallaecio, line 41:
```suggestion
    conditions given in the settings. The purpose is to skip a monitor 
    based on stat values or any custom function. A scenario could be skipping
    the Field Coverage Monitor when a spider produced no items. Following is a
    code block of examples of how we can configure the skip rules in settings.  
```

spidermon/contrib/scrapy/monitors/base.py
```diff
@@ -1,18 +1,93 @@
...
+                "SPIDERMON_FIELD_COVERAGE_RULES": {
+                    "dict/quote": 1,
+                    "dict/author": 1,
+                },
+                "SPIDERMON_MONITOR_SKIPPING_RULES": {
+                    "Field Coverage Monitor": [["item_scraped_count", "==", 0]],
+                }
+            }
+
+    Example#2: skip rules based on custom function
```
Gallaecio, line 57:
```suggestion
    Example #2: skip rules based on a custom function
```

spidermon/contrib/scrapy/monitors/base.py
```diff
@@ -1,18 +1,93 @@
...
+                    "dict/author": 1,
+                },
+                "SPIDERMON_MONITOR_SKIPPING_RULES": {
+                    "Field Coverage Monitor": [skip_function],
+                }
+            }
+    """
+
+    def check_if_skip_rule_met(self):
+        if hasattr(self, "skip_rules") and self.skip_rules.get(self.name.split("/")[0]):
```
Gallaecio, line 78: Why the `split` in `self.name.split("/")[0]`?

Also, isn’t `Monitor.name` optional? And, if so, don’t we need to account for it being `None` here? (even if it is only to show a better error message).

spidermon/contrib/scrapy/monitors/suites.py
```diff
@@ -38,6 +38,30 @@ class SpiderCloseMonitorSuite(MonitorSuite):
...
+            monitors_finished_actions,
+            monitors_passed_actions,
+            monitors_failed_actions,
+            order,
+            crawler,
+        )
+        if dict(crawler.settings).get("SPIDERMON_MONITOR_SKIPPING_RULES"):
+            skip_rules = crawler.settings.get("SPIDERMON_MONITOR_SKIPPING_RULES")
+            for monitor in self.monitors:
+                monitor.skip_rules = skip_rules
```
Gallaecio, line 63: Why are we assigning the dict of monitors to every monitor. Since rules are based on monitor name, couldn’t we filter them at this point, and only assign to each monitor its own rules, and not set rules altogether on monitors for which there are no rules?

shafiq-muhammad: At this point, monitor's properties are not set (we can't get the monitor.name or monitor.monitor_name)

Gallaecio: I see :sweat:.

I was hoping monitor had access to the Scrapy crawler of settings, but I see that’ implementing that would me a major change, and I don’t think it makes sense to make it as part of this.

Given the current code base, I think your approach is the cleanest it can be. Great work!

Gallaecio (approved):

PR #389, opened 2023-02-28
Adding step to log spidermon related settings

### LINKED ISSUE #388: It can be difficult to confirm current spidermon thresholds for monitors
We recevied from feedback that it is common for users to have difficulty confirming what a spider's current monitor thresholds are. This is due to the myriad ways these can be set:

- In a project's settings.py file.
- In a spider's custom_settings attribute.
- In the project settings tab on the zyte dashboard.
- In the spider settigns tab on the zyte dashboard.

The suggestion we have received is to include logs that state the current thresholds to make it clear to a user what targets the monitors are following.

further-reading:
Fixes #388


further-reading:
Tests are failing because most of the tests seem to use an empty Class as a MockSpider, so have to convert all these mocks into MagicMocks.

conftest.py
```diff
@@ -16,7 +16,7 @@ def _crawler(extended_settings={}):
         }
         settings.update(extended_settings)
         crawler = Crawler(Spider, settings=settings)
-        crawler.spider = Spider("dummy")
+        crawler.spider = Spider.from_crawler(crawler, name="dummy")
```
Gallaecio, line 19: If you want to make things more future-proof, you can use https://github.com/scrapy/scrapy/blob/8fbebfa943c3352f5ba49f46531a6ccdd0b52b60/scrapy/utils/misc.py#L143

Gallaecio (approved):

spidermon/contrib/scrapy/runners.py
```diff
@@ -73,6 +73,11 @@ def add_action_error(self, action, error):
         super().add_action_error(action, error)
         self.write_item_result(action)
 
+    def declare_settings(self):
+        for key, value in self.spider.crawler.settings.items():
+            if "SPIDERMON" in key:
+                self.log_info(f"{key}: {value}")
```
VMRuiz, line 79: We should consider logging it as debug instead of info.

VMRuiz (commented):
Is this code executed once per Scrapy job, per test suite, or per monitor? Additionally, it would be useful to have the ability to turn this feature on or off

further-reading:
> Is this code executed once per Scrapy job, per test suite, or per monitor? Additionally, it would be useful to have the ability to turn this feature on or off

It's per test suite.

spidermon/contrib/scrapy/runners.py
further-reading: I think info fits the use case better - this data is useful even if the application is running fine. The underlying issue is the myriad places it can be set makes it hard for a user to get this data at all.

further-reading: > If I have multiples suites, why would I want to have the settings printed several times?

It was a limitation with the reporting logic spidermon uses - to handle it in the same way other spidermon reports are made with minimal impact to the code base it ended up triggering per test suite.

For the rest of the feedback you raise an interesting point. More succinctly put - the lack of clarity of settings at run time is a general scrapy problem, not a problem specific to spidermon. Scrapy users in general have this issue regardless of if they are spidermon users. 

With this in mind - do you feel this issue should be closed? Either a seperate extention could be made (which also fixes the redundant reporting + outputs it in a json format) or an option for scrapy to confirm all settings could be more appropriate ways to handle the root cause.

further-reading: I've gone ahead and created a seperate scrapy extention for general logging of settings that covers this use case. This PR and issue can be closed if we don't want a spidermon specific implemention.

PR #391, opened 2023-02-28
Support for list Field Coverage

### LINKED ISSUE #390: Add support for lists of dictionaries in field coverage rules
At the moment is possible to track field coverage on dictionaries and Items, supporting flat and nested fields.
A use case was raised where it would be useful to track coverage on fields that are lists of dictionaries. The idea is to provide support for this use case by improving the FieldCoverageMonitor rules.

mrwbarg:
Closes #390.

When a field scraped by a spider is a list containing objects, there's no way to set thresholds for those fields. This PR adds support for correctly couting and calculating the coverage for those types of fields both at the top level of the item and inside nested structures.


spidermon/contrib/scrapy/extensions.py
```diff
@@ -149,6 +149,14 @@ def _count_item(self, item, skip_none_values, item_count_stat=None):
                 self._count_item(value, skip_none_values, field_item_count_stat)
                 continue
 
+            if isinstance(value, list):
+                items_count_stat = f"{field_item_count_stat}/_items"
+                for list_item in value:
+                    self.crawler.stats.inc_value(items_count_stat)
+                    if isinstance(list_item, dict):
```
VMRuiz, line 156: Should we also count nested lists?

VMRuiz (commented):
I think this feature is interesting, but I'm concerned about the impact that iterating over each array of an item might have on performance. Have you tested this patch by running some Scrapy jobs with complex or larger items to see if they perform well compared to the basic version?

mrwbarg:
> I think this feature is interesting, but I'm concerned about the impact that iterating over each array of an item might have on performance. Have you tested this patch by running some Scrapy jobs with complex or larger items to see if they perform well compared to the basic version?

I haven't run any test jobs. But this will definitely impact performance. For example, if we assume all items in a job have one field which is a list of `m` objects and the job scraped `n` items. The overall complexity for this would be `O(mn)` while previously it would have been just `O(n)`. If `m` is big enough, the impact will definitely be felt. Of course, this gets worse if there are more lists inside the objects that are on the topmost list (this can be avoided though).

All the ways I can think of doing this (using a `Counter` for example) would still require the entire list to be traversed for each yielded item. Do you have any implementations suggestions that might mitigate this issue?

spidermon/contrib/scrapy/extensions.py
mrwbarg: If we don't, this could definitely improve performance.

mrwbarg:
Updated it so now it is enabled through a setting and the coverage nesting levels can be set.

Gallaecio:
If I understand correctly, this is disabled by default. If so, we can leave it up to users to decide whether they are willing to enable this at the cost of the corresponding performance hit. Maybe mentioning the potential performance hit in the setting documentation is enough.

mrwbarg:
> If I understand correctly, this is disabled by default. If so, we can leave it up to users to decide whether they are willing to enable this at the cost of the corresponding performance hit. Maybe mentioning the potential performance hit in the setting documentation is enough.

yup, you're correct

curita:
Hi all! To make sure, is anything blocking this PR from getting approved? There's a comment in the docs from this PR about the performance impact of enabling this setting. I think that's covered now.

Gallaecio (approved):
:heavy_check_mark: from me, I did not realize the new docs already mentioned performance.

VMRuiz (approved):

PR #392, opened 2023-03-03
Update contributing guidelines to include pre-commit install

VMRuiz:


CONTRIBUTING.rst
```diff
@@ -85,10 +85,14 @@ Ready to contribute? Here's how to set up `spidermon` for local development.
     $ pip install -r requirements-docs.txt
     $ tox
 
+#. Install the git hook scripts from pre-commit::
+
+    $ pre-commit install
+
 #. Make sure that your code is correctly formatted using `black`_ . No code will
    be merged without this step::
```
kmike, line 93: I think this step is no longer needed, pre-commit hook should be handling it

requirements.txt
```diff
@@ -14,4 +14,5 @@ black
 pytest-mock
 scrapinghub
 scrapinghub-entrypoint-scrapy
-itemadapter
\ No newline at end of file
+itemadapter
+pre-commit
```
kmike, line 18: I'm not sure how requirements.txt are organized, but probably it'd be better to make an explicit requirement-dev.txt file, as not all of these requirements are needed to run spidermon.

Also, it seems some of the requirements which are here are not in setup.py, e.g. itemadapter is not in setup.py install_requires (it's in test_requires for some reason).

CONTRIBUTING.rst
VMRuiz: Afaik, pre-commit only check that it was done before allowing the commit. Developers are still responsible for doing it themselves.

requirements.txt
VMRuiz: Yeah, the package requirements are weirdly structured, we could open a new issue to fix that

CONTRIBUTING.rst
kmike: Usually we configure pre-commit to actually change the files. So, it runs black and other tools; if there are no changes, it allows to commit, but if there are changes, then it blocks the commit - but the files are changed on disk. Then, user needs to add the changed files to the git staging area, and then commit again. 

This workflow is nice, because there is no opportunity e.g. to use a different black version, there is no need to run formatters manually, but also there is a change to review the changes before making a commit.

VMRuiz: You are right. This part is now removed.

requirements.txt
kmike: Makes sense; we don't need to fix it in this PR.

Not having itemadapter in install_requires looks like a separate issue, I think it's a bug.

kmike (approved):
👍

PR #397, opened 2023-04-24
Allow kwargs on Slack API.

### LINKED ISSUE #293: Allow pass extra optional arguments to Slack API

mrwbarg:
Closes #293.

Allows keyword args to be passed to the slack API.


spidermon/contrib/actions/slack/__init__.py
```diff
@@ -171,6 +183,7 @@ class SendSlackMessage(ActionWithTemplates):
     include_message = True
     include_attachments = True
     fake = False
+    kwargs = {}
```
Gallaecio, line 186: It may be best not to assign `{}` as a default value. Do we even need this attribute here?

mrwbarg: Because it's assigned in the constructor and used in run_action I think it needs to be stored somewhere until it's used. What would you suggest as a default value? I'd thought about None but then, when it needs to be used (line 257) I would just do **self.kwargs if self.kwargs else {} which would accomplish the same I think.

mrwbarg: Sounds good! Fixed.

Gallaecio (approved):

VMRuiz (approved):

PR #398, opened 2023-05-30
Change supported versions

### LINKED ISSUE #394: Update supported python versions
- Add support for python3.11
- Drop support for python3.6 - security support finished on 23 Dec 2021 -
- Maybe drop support for python3.7 - security support will finish on 27 Jun 2023 -

mrwbarg:
Closes #394.

Updating supported python versions.


mrwbarg:
Seems like tests were failing because in Python 3.11 `re.compile('[0-9]++')` is valid (which is not true). I added a bypass for now and opened an issue: https://github.com/python/cpython/issues/105121

curita:
@mrwbarg I think now "++" is a valid regex, looking at the updated `re` docs: https://docs.python.org/3/library/re.html. It was introduced in 3.11. But changing that test case to any other invalid regex should do it, which you already covered 👍

VMRuiz (approved):

wRAR (approved):

Issue #370, opened 2022-11-29
Schematics seems to no longer be supported.


paulvav:
Hi Spidermon team, 

Was looking over the schematics project and noticed it looks like it is no longer maintained. I note that it doesn't support Python 3.10 and doesn't appear to have any code changes for over a year.
See also:
https://github.com/schematics/schematics/issues/628

Going forward should the recommendation be to use jsonschema exclusively for item validation? Or will another validation library be considered - maybe https://github.com/pydantic/pydantic ?

rennerocha:
Hello @paulvav ,

`schematics` looks dead to me as well and I have no good hope that they will merge this 3.10 PR soon (and 3.11 is already there for a while).

From my point of view, considering the projects that I saw when I worked at Zyte (ex-Scrapinghub) which is the company that sponsor Spidermon, I don't remember any of them that uses schematics. jsonschema is the standard for item validation there. Perhaps outside Zyte this could be the same?

I would suggest to deprecate schematics (adding a Deprecation warning for anyone using it) in the next release (1.18.0) and them remove it from Spidermon in 1.19.0. This would make the item validation pipeline much cleaner too.

Any thoughts on that @VMRuiz @Gallaecio ?

VMRuiz:
Hello @paulvav , @rennerocha , 

Schematics looks abandoned indeed. [Even the doc warms is _currently somewhat_ outdated](https://schematics.readthedocs.io/en/latest/). So I'm good with deprecating this eventually. 

If there is enough interest on other validation tools we could think of adding them. 
But that is a different topic that doesn't need to wait for this one to resolve.

HaddadJoe:
@VMRuiz schematics is broken starting python 3.10 and beyond. Tightly coupling spidermon with a dead project (latest update from 2021) is not a great idea for spidermon moving forward. this is preventing us from upgrading our spider's python version. Would you be open for PRs that drop support of schematics in favor of other validation libraries? If yes, any preference on which solutions to adopt?

VMRuiz:
Hello @HaddadJoe, thanks for you interest on the topic!

We are open to PR both for deprecating schematics and to add new validations libraries. 
They don't necessarily needs to be done simultaneously

raphapassini:
I'm working on that. I'll have a PR ready soon.

PR #399, opened 2023-06-05
Deprecate schematics

mrwbarg:
Addresses parts of #370.

Deprecate the use of schematics as it's unlikely to receive support after Python 3.11.
This PR only removes schematics but doesn't replace it with another library (as we already have support for jsonschema).
A future PR can add support to another library such as Pydantic for example.


VMRuiz:
I'm in favor of deprecating this, but should we give some warnings first? We can add some warning messages saying that schematics support will be drop in futures versions.

VMRuiz (approved):

PR #402, opened 2023-06-29
Periodic item count monitor

### LINKED ISSUE #324: Built-in periodic monitor to check if the amount of items increases
Spidermon could have a periodic monitor that validates that the amount of items (tracked in the stat `item_scraped_count`) increases after every check.

This will alert us of jobs that are stuck without returning any new items but aren't necessarily throwing any issues.

There could be a threshold to indicate by how much percentage the item count is expected to increase, to alert us for jobs that are working slower than expected as well. There could be two possible settings I imagine too, one absolute one (meaning it will check that there's at least x new items after every check) and a relative one (checking the item increase is at least y%), with the possibility of defining one or both.

We have implemented something similar in one of our [projects](https://bitbucket.org/scrapinghub/%7B47f03fcb-d09e-4046-9350-0b9277e06a3c%7D/pull-requests/1270) [private], that could be taken as a starting point after stripping the project related code.

ogiaquino:
Closes #324 

Add a monitor suite to periodically monitor increase in item_scraped_count.
Increase in item count can be set by adding SPIDERMON_ITEM_COUNT_INCREASE in project settings.

The value for SPIDERMON_ITEM_COUNT_INCREASE can be integer to check for x amount increase in items.
The value for SPIDERMON_ITEM_COUNT_INCREASE can be float to check in percentage increase of items.

Below is a sample to enable this monitor suite:

```
SPIDERMON_ITEM_COUNT_INCREASE = 100

SPIDERMON_PERIODIC_MONITORS = {
    'spidermon.contrib.scrapy.monitors.PeriodicItemCountMonitorSuite': 2,
}
```


spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -578,3 +579,31 @@ def get_threshold(self):
...
+    in percentage increase of items.
+    """
+
+    stat_name = "item_scraped_count"
+    threshold_setting = "SPIDERMON_ITEM_COUNT_INCREASE"
+    assert_type = ">="
+
+    def get_threshold(self):
+        crawler = self.data.crawler
+        prev_item_scraped_count = self.stats.get("prev_item_scraped_count", 0)
```
VMRuiz, line 600: Most monitor don't assume a default value now and raise and error if used without it being configured. Maybe we should follow this approach here too?

VMRuiz: Example : https://spidermon.readthedocs.io/en/latest/monitors.html#spidermon.contrib.scrapy.monitors.monitors.ErrorCountMonitor

VMRuiz (commented):
Small details about default value. Overall looks good to me. Great Idea, thanks!

spidermon/contrib/scrapy/monitors/monitors.py
ogiaquino: Thanks for the feedback @VMRuiz I've added raise NotConfigured if SPIDERMON_ITEM_COUNT_INCREASE is not configure.

VMRuiz (approved):

further-reading (approved):
LGTM :)

PR #407, opened 2023-07-10
Refactor requirements

### LINKED ISSUE #393: Dependencies needs some refactoring
There is some confusion on where each dependency should go. For example, ` requirements.txt` contains dependencies both for Spidermon and auxilar development tools like `black` and `pre-commit`.

There is a `requirements-docs.txt` but not a `requirements-dev.txt`

Then, `setup.py` contains `extra_require` both for Spidermon features - `monitoring`, `validation`, and for Spidermon development `tests`, `docs` . At the sametime, some dependencies like `itemadapter` are missing.

mrwbarg:
Closes #393.
Organizing dependencies.


VMRuiz (changes requested):
I'm good with the current approach but I think it was suggested to get rid of `requirements.txt` files and just leave pre-commit & tox to fetch all dependencies from the setup.py file.

Some examples:

- https://github.com/scrapy/scrapy/blob/master/tox.ini
- https://github.com/scrapy-plugins/scrapy-zyte-api/blob/main/tox.ini

VMRuiz:
Could you also please update https://github.com/scrapinghub/spidermon/blob/master/CONTRIBUTING.rst to replace references to requirements.txt files and substitute it with just installing tox ?

docs/requirements-docs.txt
```diff
@@ -0,0 +1,3 @@
+sphinx
+sphinx-rtd-theme
```
kmike, line 2: It can be a good practice to pin the package versions here.

setup.py
```diff
@@ -29,24 +14,24 @@
...
-            "boto",
-            "boto3",
-            "premailer",
-            "sentry-sdk",
-        ],
-        # Tools to run the tests
-        "tests": test_requirements,
-        # Tools to build and publish the documentation
-        "docs": ["sphinx", "sphinx-rtd-theme", "s3cmd"],
+        "monitoring": [],
```
kmike, line 34: Would it make sense to move packags like slack_sdk, sentry_sdk and boto here? Are they optional dependencies of spidermon, or are they all required for spidermon to work?

tox.ini
```diff
@@ -3,19 +3,19 @@ envlist = min,py39,py310,py311
...
+deps =
+    pytest
+    pytest-cov
+    pytest-mock
 commands = pytest -s --cov=spidermon --cov-report= {posargs:tests}
 
 [testenv:min]
 basepython = python3.8
-deps =
-    jsonschema[format]==3.2.0
```
kmike, line 14: Why is this line removed? I think the intention was to pin to a min package version, and test against it; in other environments a lates vertsion is going to be installed. It could make sense to pin other packages here as well (like scrapy), if a minimum version requirement is known.

setup.py
VMRuiz: They are not required by Spidermon itself but by different actions:
- boto is required by `actions/reports/s3.py`
- boto3 is required by `actions/email/ses.py`
- sentry_sdk is required by `actions/sentry/__init__.py`
- slack_sdk is required by `actions/slack/__init__.py`

Before this change, most-common user case of spidermon was to install everything with `spidermon[monitoring]`. The idea was to make this the default installing by moving all the packages into `install_requires` and keep `monitoring` empty to be backward compatible. 

Alternatively, we could move this packages into a new section named `actions` but that wouldn't be backwards compatible and may be quite problematic for most people.

tox.ini
VMRuiz: My bad, I saw it was already pinned in `install_requires` but I didn't noticed that it was `>=` instead of `==`

VMRuiz (approved):

VMRuiz (commented):
Setting my review to neutral

setup.py
wRAR: What about the reqs that were moved from `tests_require` / `spidermon[tests]` to `install_requires`?

VMRuiz: If I'm not mistake those would be: 

- itemadapter: [[contrib/scrapy/pipelines.py](https://github.com/scrapinghub/spidermon/blob/e6509dd0523acded0d6cf7dd3422a9581a99815a/spidermon/contrib/scrapy/pipelines.py#L4)], [[contrib/scrapy/extensions.py](https://github.com/scrapinghub/spidermon/blob/e6509dd0523acded0d6cf7dd3422a9581a99815a/spidermon/contrib/scrapy/extensions.py#L5)], 
- requests [[contrib/actions/telegram/__init__.py](https://github.com/scrapinghub/spidermon/blob/e6509dd0523acded0d6cf7dd3422a9581a99815a/spidermon/contrib/actions/telegram/__init__.py#L3)],[[contrib/actions/discord/__init__.py](https://github.com/scrapinghub/spidermon/blob/e6509dd0523acded0d6cf7dd3422a9581a99815a/spidermon/contrib/actions/discord/__init__.py#L5)]: 
- scrapinghub-entrypoint-scrapy (sh_scrapy) [[contrib/stats/statscollectors/sc_collections.py](https://github.com/scrapinghub/spidermon/blob/e6509dd0523acded0d6cf7dd3422a9581a99815a/spidermon/contrib/stats/statscollectors/sc_collections.py#L6)]
- twisted [[contrib/scrapy/extensions.py](https://github.com/scrapinghub/spidermon/blob/e6509dd0523acded0d6cf7dd3422a9581a99815a/spidermon/contrib/scrapy/extensions.py#L4)]

itemadapter, requests and sh_scrapy are required by different parts of the code. I guess they were not previously added by mistake, because they are already present on most Scrapy stacks anyway. 

The only package that seems redundant to me is twisted which I think we should drop as is already covered by `scrapy`.

kmike: I assume the dependencies like boto or slack-sdk are all optional, and most users don't use all of them. 

Not installing them by default minimizes scenarios when some of these packages don't install for some reason for the users - e.g. a package drops support for some Python version which spidermon still supports (and which an user is using), or requirements of some package conflict with requirements which user's project has. 

If these packages are not optional, then in such scenario spidemon installation is broken for users who don't even use these features.

That said, if based on your experience with these particular packages you found it unlikely they break, it looks fine to install them by default.

So, I'm approving the PR, and let you folks decide if the concern is worth addressing, or if addressing it just complicates things.

If it's something which should be addressed, probably the easiest way to go is to move all optional requirements to some section like `pip install spidermon[contrib]` or `pip install spidermon[all]`, keep only the required ones in default install_requires, and also document `pip install spidermon[all]` as the way to install spidermon.

kmike (approved):

setup.py
VMRuiz: I'm worried about adding a new optional requirements like `contrib` all `all` after we already have `monitors` in place that is used by default by most users.

VMRuiz:
I have updated the PR with some changes: 

- Remove optional dependencies from install_requires and put them as extras
- Keep extra `monitoring` naming for backward compatibility. We can refactor this later to `all` or `contrib` if needed. 
- Split tox configuration between `testenv` (_all tests + all extras_) and `testenv:base` (_core tests + no extras except Scrapy and Jinja2_). The reason for adding Scrapy and Jinja2 to testenv:base is because default `conftest.py` and others top-level tests contains dependencies on them. We should fix that in a following PR.

Gallaecio (approved):

PR #413, opened 2023-08-25
Add SNS Action for Spidermon Notifications

felipdc:
### Summary:
This pull request introduces a new action that allows Spidermon to send notifications via AWS Simple Notification Service (SNS). This provides users with an alternative way to receive notifications and integrate Spidermon with other AWS services or third-party applications through SNS.

### Changes:
- Added `SendSNSNotification` class to handle SNS notifications.
- Provided documentation on how to set up and use the SNS action.

### Benefits:
- Offers more flexibility for users who are heavily invested in the AWS ecosystem.
- Allows for easy integration with other AWS services like Lambda, SQS, etc., by leveraging SNS topics.

### How to Test:
1. Set up AWS credentials and an SNS topic as described in the documentation.
2. Configure Spidermon to use the new SNS action.
3. Trigger a Spidermon notification to see it being sent to the specified SNS topic.

Looking forward to feedback and any additional changes required.


docs/source/actions/sns-action.rst
```diff
@@ -0,0 +1,52 @@
...
+
+This action allows you to send custom notifications to an AWS `Simple Notification Service (SNS)`_ topic when your monitor suites finish their execution.
+
+To use this action, you need to provide the AWS credentials and the SNS topic ARN in your ``settings.py`` file as follows:
+
+.. code-block:: python
+
+    # settings.py
+    SPIDERMON_SNS_TOPIC_ARN = '<SNS_TOPIC_ARN>'
+    SPIDERMON_AWS_ACCESS_KEY_ID = '<AWS_ACCESS_KEY>'
```
VMRuiz, line 12: In the future, we may want to refactor all Actions using `SPIDERMON_AWS_ACCESS_KEY_ID` to be able to  define their own `SPIDERMON_{ACTION}_AWS_ACCESS_KEY_ID` or to use `SPIDERMON_AWS_ACCESS_KEY_ID` as fallback.

That would help in case you want to use different keys for each type of service. 

cc. @Gallaecio , @fcanobrash  @curita

spidermon/contrib/actions/sns/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+    @classmethod
+    def from_crawler_kwargs(cls, crawler):
+        return {
+            "topic_arn": crawler.settings.get("SPIDERMON_SNS_TOPIC_ARN"),
+            "aws_access_key": crawler.settings.get("SPIDERMON_AWS_ACCESS_KEY_ID"),
+            "aws_secret_key": crawler.settings.get("SPIDERMON_AWS_SECRET_ACCESS_KEY"),
+        }
+
+
+class SendSNSNotification(BaseSNSNotification):
```
VMRuiz, line 76: What's the difference with `BaseSNSNotification` ? Why is this class not used anywhere else?

docs/source/actions/sns-action.rst
Gallaecio: For such cases, users should probably be encouraged to create their own action classes and set those parameters through `super().__init__()`. I imagine there could be scenarios where you want to have 2 instances of the same action with different settings, so you might need to do that anyway.

But I guess an action-specific setting that uses the global one as fallback would indeed minimize scenarios requiring subclassing, so +1 from me.

rennerocha: In my opinion we should keep having only one place with AWS credentials to keep the possible settings available manageable. It is more likely that someone will have the same keys for any AWS integration.

`SPIDERMON_AWS_ACCESS_KEY_ID` and `SPIDERMON_AWS_SECRET_ACCESS_KEY` are common keys that any AWS integration will require, and in my experience, it is very unlikely that they will have different ones for different services (and if so, they can create their own custom actions)

I would change `SPIDERMON_SNS_TOPIC_ARN` to `SPIDERMON_AWS_SNS_TOPIC_ARN` to make it more clear that we are setting a AWS setting to the project.

This follows the patter we already have for `SPIDERMON_AWS_REGION_NAME` and `SPIDERMON_AWS_RETURN_PATH`

rennerocha:
**thought** This is not related to this PR specifically, but maybe it would be nice, if we move any action that it is AWS related to a `aws` module. This would keep the code more organized and easier to locate.

So we could have:
```
spidermon/contrib/actions/aws/__init__.py
spidermon/contrib/actions/aws/sns.py
spidermon/contrib/actions/aws/ses.py  # for Amazon SES email
```
Maybe we could discuss this in a separate issue anyway.

VMRuiz (approved):
LGTM

PR #415, opened 2023-08-30
Fix ItemValidationPipeline using __setitem__ without ItemAdapter

VMRuiz:


VMRuiz:
I don't think we need to conver to dict [here](https://github.com/scrapinghub/spidermon/pull/415/files#diff-7937ac85a30630fe837b9c133f4459ee590680bb5dfce72775db6005f2b45f51R109
) and [here](https://github.com/scrapinghub/spidermon/pull/415/files#diff-7937ac85a30630fe837b9c133f4459ee590680bb5dfce72775db6005f2b45f51R111) but I'm keeping the same logic for now

spidermon/contrib/scrapy/pipelines.py
```diff
@@ -103,15 +103,16 @@ def process_item(self, item, _):
             # No validators match this specific item type
             return item
 
-        data = self._convert_item_to_dict(item)
+        item_adapter = ItemAdapter(item)
+        item_dict = item_adapter.asdict()
         self.stats.add_item()
-        self.stats.add_fields(len(list(data.keys())))
+        self.stats.add_fields(len(list(item_dict.keys())))
```
rennerocha, line 109: **nitpick** we don't need to convert to `list`. We can use: `self.stats.add_fields(len(item_dict.keys()))`

PR #417, opened 2023-08-31
Add support for nested fields in SPIDERMON_VALIDATION_ERRORS_FIELD

VMRuiz:
Allows to use a nested field as validation field. Nested fields are supported by using `.` separator:

```python
# settings.py
SPIDERMON_VALIDATION_ERRORS_FIELD = "top_level.second_level._validation"
```

### DIFF
--- a/docs/source/item-validation.rst
+++ b/docs/source/item-validation.rst
@@ -102,7 +102,13 @@ SPIDERMON_VALIDATION_ERRORS_FIELD
 Default: ``_validation``
 
 The name of the field added to the item when a validation error happens and
-`SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS`_ is enabled.
+`SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS`_ is enabled. Nested fields are supported by using `.` separator:
+
+.. code-block:: python
+
+    # settings.py
+    SPIDERMON_VALIDATION_ERRORS_FIELD = "top_level.second_level._validation"
+
 
 SPIDERMON_VALIDATION_SCHEMAS
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


docs/source/item-validation.rst
```diff
@@ -102,7 +102,13 @@ SPIDERMON_VALIDATION_ERRORS_FIELD
 Default: ``_validation``
 
 The name of the field added to the item when a validation error happens and
-`SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS`_ is enabled.
+`SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS`_ is enabled. Nested fields are supported by using `.` separator:
```
Gallaecio, line 105:
```suggestion
`SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS`_ is enabled. Nested fields are supported by using ``.`` as a separator:
```

or 

```suggestion
`SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS`_ is enabled. Nested fields are supported by using the ``.`` separator:
```

I wonder if we should acknowledge the lack of support for fields with dots (which I assume are possible), so that users don’t waste time trying escape mechanisms.

Gallaecio (approved):

docs/source/item-validation.rst
VMRuiz: That's a good question. I thought dots where a good idea because you couldn't define attributes with dots on them. However, it's actually possible to do it with dictionaries `{"dotted.key": True}`. That means there is a possibility that these new functionality breaks it for such people. For example:

```
# SPIDERMON_VALIDATION_ERRORS_FIELD = "foo._validation"
a = {"foo._validation" : [] }
```

Would have worked previously, but now it would raise `KeyError` because `foo` is not defined.

Is there a better alternative? Should we allow to escape dots someway? Is this so improbable that is not worth support it?

VMRuiz: JMESPath is basically dots with nice features, isn't? :smile:

Gallaecio: Nice features including being able to escape dots. But I guess it does break things the same way, only that the solution rather than a setting to disable nesting would be changing the expression to escape dots.

VMRuiz: We decided there is little to none risk in introducing this as it is.

PR #421, opened 2023-09-25
Adds limit for nested dict stats computation

### LINKED ISSUE #416: Support restricting the coverage monitor fields
## Background

Currently, the coverage monitor tracks and reports the coverage of all fields, including nested fields (i.e., keys inside dictionary values specifically). It follows all nested field levels.

This sometimes isn't desired, especially for dictionaries that have non-standardized keys. The coverage tracking can stay the same if there's a fixed schema for those dictionaries or too many nested levels. This is particularly troublesome when running spiders in ScrapyCloud, as there's a hard limit on the stats's storage size.

## Alternatives

1. New setting to skip the coverage tracking on fields defined by name or regex. It should support nested field definitions as well.
2. Support limiting the nested levels of coverage, like `SPIDERMON_LIST_FIELDS_COVERAGE_LEVELS`, but for dictionaries.
3. Allow the possibility of storing the coverage tracking stats somewhere else. For example, each `<key: field count>` pair could be stored as a separate entry in a ScrapyCloud collection to avoid triggering its size limitation.

All of them could be viable and coexist as they address different parts of the problem.

Let me know what you think!

mrwbarg:
Closes #416 
Adds a setting to limit how deep the coverage is computed for nested dicts.


curita:
@Gallaecio @rennerocha could you check on this PR? It's ready to be reviewed.

Gallaecio (approved):
It is OK by me. However, I wonder if we should consider making the default -1 for no-limit, and make 0 cause this:

```python
      'spidermon_item_scraped_count/dict': 2
      'spidermon_item_scraped_count/dict/field1': 2
      'spidermon_item_scraped_count/dict/field2': 2
      'spidermon_item_scraped_count/dict/field3': 1
      'spidermon_item_scraped_count/dict/field4': 1
```

VMRuiz:
Can we close this in favor of https://github.com/scrapinghub/spidermon/pull/433 ?

PR #424, opened 2023-10-15
fix: (423-fix_time_zone) - making timezone aware if needed

### LINKED ISSUE #423: Add handling for datetime aware start time in PeriodicExecutionTimeMonitor
In the current version of scrapy, the code below breaks due to spider start time being timezone aware
https://github.com/scrapinghub/spidermon/blob/master/spidermon/contrib/scrapy/monitors/monitors.py#L497

```python
now = datetime.datetime.utcnow()
start_time = self.data.stats.get("start_time")
duration = now - start_time
```

It currently throws the following error:

`TypeError: can't subtract offset-native and offset-aware datetimes`

Backwards compatible Fix: add try/except to add utc timezone to now timestamp when error thrown.

further-reading:
Closes #423 
Error handling adds timezone if needed.


wRAR:
Tests fail because they expect `self.crawler.stats` which is None, I assume it's because of the Scrapy 2.11 change to the crawler initialization. One way to work around this in tests is to assign `MemoryStatsCollector(crawler)` to it manually if it's None (if tests don't expect any other components of the Crawler instance to be initialized), like in https://github.com/scrapinghub/scrapy-poet/pull/165

curita:
The CI issues have been fixed 🙏 Could you rebase/merge the latest changes from `master`? @further-reading

wRAR:
I'm not sure but I think the test isn't related to the code being tested? I assume it increased the coverage for some of the added lines but not because it tests the added code.

VMRuiz:
Fixed at https://github.com/scrapinghub/spidermon/pull/439

Issue #425, opened 2023-10-25
Change format of content of _validation field


rennerocha:
If a validation error is found and [SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS](https://spidermon.readthedocs.io/en/1.20.0/item-validation.html#spidermon-validation-add-errors-to-items) setting is `True`, we populate the `_validation` field with the errors.

The content of this field is the string representation of a Python defaultdict such as:
` '_validation': defaultdict(<class 'list'>, {'author_url': ['Invalid URL']})`

To make this field valid as JSON, we should have it as a list of objects such as:
` '_validation': [ {'author_url': ['Invalid URL'] }]`

This change will improve the data returned, making it possible to be read as a JSON.

rochamatcomp:
Can I take this issue? I have created this pull request #431.

Laerte:
Can we close this issue since was implemented? cc: @Gallaecio @VMRuiz

PR #431, opened 2024-02-08
Change format of content of _validation field (#425)

rochamatcomp:
The content of _validation field must to be the string representation of a Python dict instead defaultdict when SPIDERMON_VALIDATION_ADD_ERRORS_TO_ITEMS setting is True.

### DIFF
--- a/docs/source/getting-started.rst
+++ b/docs/source/getting-started.rst
@@ -478,8 +478,7 @@ The resulted item will look like this:
 .. code-block:: js
 
     {
-        '_validation': defaultdict(
-            <class 'list'>, {'author_url': ['Invalid URL']}),
+        '_validation': {'author_url': ['Invalid URL']},
          'author': 'Mark Twain',
          'author_url': 'not_a_valid_url',
          'quote': 'Never tell the truth to people who are not worthy of it.',
--- a/docs/source/item-validation.rst
+++ b/docs/source/item-validation.rst
@@ -81,7 +81,7 @@ You can change the name of the field by assigning a name to `SPIDERMON_VALIDATIO
 .. code-block:: js
 
     {
-        '_validation': defaultdict(<class 'list'>, {'author_url': ['Invalid URL']}),
+        '_validation': {'author_url': ['Invalid URL']},
         'author': 'C.S. Lewis',
         'author_url': 'invalid_url',
         'quote': 'Some day you will be old enough to start reading fairy tales '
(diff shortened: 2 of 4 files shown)


VMRuiz:
Thanks @rochamatcomp for the PR!

@curita  @fcanobrash  Are you aware of any other tool that may need to be updated to adapt to this new format?

rochamatcomp:
The two failed tests are fixed in #432.

VMRuiz:
> > Thanks @rochamatcomp for the PR!
> > @curita  @fcanobrash Are you aware of any other tool that may need to be updated to adapt to this new format?
> 
> @VMRuiz given that a `defaultdict` is a subclass of `dict`, anything you were able to do with the content of `_validation` key should work fine, except if we try to access a key that is missing, which is not an expected use case. Generally in the user's code we don´t know the content of this key and just read it as it is (not trying to check for different values).

Notice that [in your issue](https://github.com/scrapinghub/spidermon/issues/425)  you said: 

> To make this field valid as JSON, we should have it as a list of objects such as:
>  '_validation': [ {'author_url': ['Invalid URL'] }]
> 

which would require to convert `_validation` type from `defaultdict` to `list`. However, the current solution converts from `defaultdict` to `dict`. This is way I was asking about possible conflicts. 

Is @rochamatcomp  approach what you were expecting?

rochamatcomp:
> > > Thanks @rochamatcomp for the PR!
> > > @curita  @fcanobrash Are you aware of any other tool that may need to be updated to adapt to this new format?
> > 
> > 
> > @VMRuiz given that a `defaultdict` is a subclass of `dict`, anything you were able to do with the content of `_validation` key should work fine, except if we try to access a key that is missing, which is not an expected use case. Generally in the user's code we don´t know the content of this key and just read it as it is (not trying to check for different values).
> 
> Notice that [in your issue](https://github.com/scrapinghub/spidermon/issues/425) you said:
> 
> > To make this field valid as JSON, we should have it as a list of objects such as:
> > '_validation': [ {'author_url': ['Invalid URL'] }]
> 
> which would require to convert `_validation` type from `defaultdict` to `list`. However, the current solution converts from `defaultdict` to `dict`. This is way I was asking about possible conflicts.
> 
> Is @rochamatcomp approach what you were expecting?

I'm expecting change from `defaultdict(list, {'_validation': [{'author_url': ['Invalid URL']}]})` to `{'_validation': [{'author_url': ['Invalid URL']}]}`

The list inside the default/dict don't change.

VMRuiz:
Sorry, my previous comment was aim to @rennerocha

VMRuiz (approved):

Issue #427, opened 2023-12-05
Periodic item count monitor crashes if it runs before item_scraped_count exists in stats


rvandam:
Encountered this error while playing with [PeriodicItemCountMonitor](https://github.com/scrapinghub/spidermon/blob/377c4ab929a1f49bc2f78787ae2d7fa615948583/spidermon/contrib/scrapy/monitors/monitors.py#L610C7-L610C31)

```
======================================================================
ERROR: Periodic Item Count Increase Monitor/test_stat_monitor
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/vandamr/.venv/46205CF6-02DD-499A-9EB8-71D8CB0FEA73/lib/python3.11/site-packages/spidermon/contrib/scrapy/monitors/base.py", line 213, in test_stat_monitor
    threshold = self._get_threshold_value()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vandamr/.venv/46205CF6-02DD-499A-9EB8-71D8CB0FEA73/lib/python3.11/site-packages/spidermon/contrib/scrapy/monitors/base.py", line 201, in _get_threshold_value
    return self.get_threshold()
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/vandamr/.venv/46205CF6-02DD-499A-9EB8-71D8CB0FEA73/lib/python3.11/site-packages/spidermon/contrib/scrapy/monitors/monitors.py", line 639, in get_threshold
    return prev_item_scraped_count + threshold_increase
           ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

This happens in `get_threshold` if `item_scraped_count` doesn't exist yet on the first run so then `prev_item_scraped_count` gets set to None so on the second run the above error happens.

```
    def get_threshold(self):
        crawler = self.data.crawler
        prev_item_scraped_count = self.stats.get("prev_item_scraped_count", 0) # <-- on second run this is then None
        item_scraped_count = self.stats.get(self.stat_name) # <--- if this doesn't exist yet
        crawler.stats.set_value("prev_item_scraped_count", item_scraped_count) # <--- then this gets set to None 
        threshold_increase = crawler.settings.get(self.threshold_setting)
        if isinstance(threshold_increase, int):
            return prev_item_scraped_count + threshold_increase
```

[...]

PR #436, opened 2024-03-08
Avoid TypeError when running at PeriodicItemCountMonitor

VMRuiz:
Fixes https://github.com/scrapinghub/spidermon/issues/427 

However, the monitor will still won't pass until item_scraped_count is defined. I'm not sure if we should include the possibility of wanting this monitor to pass when tresshold is 0 and item_scraped_count is None.

### DIFF
--- a/spidermon/contrib/scrapy/monitors/monitors.py
+++ b/spidermon/contrib/scrapy/monitors/monitors.py
@@ -633,7 +633,7 @@ class PeriodicItemCountMonitor(BaseStatMonitor):
         crawler = self.data.crawler
         prev_item_scraped_count = self.stats.get("prev_item_scraped_count", 0)
         item_scraped_count = self.stats.get(self.stat_name)
-        crawler.stats.set_value("prev_item_scraped_count", item_scraped_count)
+        crawler.stats.set_value("prev_item_scraped_count", item_scraped_count or 0)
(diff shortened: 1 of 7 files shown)


spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -633,7 +633,7 @@ def get_threshold(self):
         crawler = self.data.crawler
         prev_item_scraped_count = self.stats.get("prev_item_scraped_count", 0)
         item_scraped_count = self.stats.get(self.stat_name)
-        crawler.stats.set_value("prev_item_scraped_count", item_scraped_count)
+        crawler.stats.set_value("prev_item_scraped_count", item_scraped_count or 0)
```
Gallaecio, line 636: :lipstick: Wouldn’t adding `0` to the earlier `get()` work? i.e. doing in line 635 the same as in line 634.

Gallaecio (approved):

spidermon/contrib/scrapy/monitors/monitors.py
VMRuiz: Yes, I wasn't sure which place was better to apply the fix. I decided to do it in 636 just for the remotely case where you may want to keep the original value of `item_scraped_count` for who knows what reason.

PR #440, opened 2024-04-04
[434] Allow to filter jobs in ZyteJobsComparisonMonitor by close_reason

shafiq-muhammad:
Adds limit for nested dict stats computation

Added a new setting `SPIDERMON_JOBS_COMPARISON_CLOSE_REASONS` to allow `ZyteJobsComparisonMonitor` to filter by the ScrapyCloud jobs with their `close_reason` stat.


spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -551,6 +556,9 @@ def run(self, result):
 
     def _get_jobs(self, states, number_of_jobs):
         tags = self._get_tags_to_filter()
+        close_reason = self.crawler.settings.getlist(
```
curita, line 559: **nitpick:** `close_reason` -> `close_reasons`

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -563,7 +571,13 @@ def _get_jobs(self, states, number_of_jobs):
             filters=dict(has_tag=tags) if tags else None,
         )
         while _jobs:
-            jobs.extend(_jobs)
+            if close_reason:
+                for job in _jobs:
+                    if job.get("close_reason") in close_reason:
+                        jobs.append(job)
+            else:
+                jobs.extend(_jobs)
```
curita, line 579: **suggestion (non-blocking):** It should be alright to use `jobs.append()` for both cases to simplify a bit the code flow.

```python
while jobs:
    for job in _jobs:
        if close_reason and job.get("close_reason") not in close_reason:
            continue
        jobs.append(job)
    ...
```

curita (changes requested):
Changes look good to me!

We should add some unittest to cover the new addition, and it would be nice to run some test job in ScrapyCloud to verify these changes (Installing Spidermon in the requierements.txt from `git+https://github.com/shafiq-muhammad/spidermon@shafiq-434` should get you these changes). After that's addressed we should be good to go 👍

VMRuiz:
Fixed at https://github.com/scrapinghub/spidermon/pull/439

aaneto:
Just a quick FYI @shafiq-muhammad @curita, I believe the pipeline is failing due to a regression from pytest 8.1.1 to 8.2.0.

My validation was:
- Creating an empty .env for the project and running `pip install tox` and `tox -e base`
- Noticing the errors, and setting pytest==8.1.1 on tox.ini
- Running `tox -e base` again, with success this time.

Hope this helps (and is correct regardless of environment).

VMRuiz:
Yes, it looks like Pytest 8.2.0 release has caused a lot of troubles according to their issues page. 
@shafiq-muhammad , Could you please pin pytest to 8.1.1 until they release a new version?

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -528,6 +529,10 @@ class ZyteJobsComparisonMonitor(BaseStatMonitor):
     You can also filter which jobs to compare based on their tags using the
     ``SPIDERMON_JOBS_COMPARISON_TAGS`` setting. Among the defined tags we consider only those
     that are also present in the current job.
+
+    You can also filter which jobs to compare based on their close reason using the
+    ``SPIDERMON_JOBS_COMPARISON_CLOSE_REASONS`` setting. The default value is ``()``,
+    which doesn't filter any job based on close_reason.
```
curita, line 535: **suggestion:** I'd maybe add an example as I think most developers might want to set this.

Something like this:

 ```suggestion
    which doesn't filter any job based on close_reason. To only consider successfully finished jobs,
    use ``("finished", ) instead.``
```

File: spidermon/contrib/scrapy/monitors/monitors.py [hunk 4]
```diff
@@ -571,9 +579,13 @@ def _get_jobs(self, states, number_of_jobs):
                 count=count,
                 has_tag=tags or None,
             )
-            total_jobs.extend(current_jobs)
 
-            if len(current_jobs) < MAX_API_COUNT or len(total_jobs) == number_of_jobs:
+            for job in current_jobs:
```
curita, line 583: **thought:** This makes sense since it seems we can't filter by close_reason in `client.spider.jobs.list()`.

I wonder if it makes sense to retrieve `MAX_API_COUNT` count jobs each time from the API and append to `total_jobs` the jobs that meet the criteria until we reach `number_of_jobs`. This should result in fewer API calls, but those calls will be larger. What do you think, @VMRuiz?

curita (changes requested):
Changes look good to me, though we might want to change how we get the jobs depending on whether we want to get fewer API calls. I'm inclined to get fewer API calls, but let's wait for Victor's opinion.

Other than that, I suggested a small documentation addition, which should be quick to include.

spidermon/contrib/scrapy/monitors/monitors.py
VMRuiz: If the default close reason is `finished` I would keep it as it is. As the most common scenario will be that you fetch the last few jobs and those will have the finished status. Even in the case where a second call is needed - due to a few unexpected failed jobs in between - it could still be faster for lower volumes than making 1 very big call.   

On the contrary, if it's expected that uncommon close reasons like `timeout` are used, then it makes sense to fetch as many jobs as possible, as they will be few or even none.

curita (approved):
Looks ready to me!

VMRuiz (approved):

Issue #442, opened 2024-04-17
issue: Wrong previous jobs count in ZyteJobsComparisonMonitor


curita:
## Background

The method to get the previous jobs used in ZyteJobsComparisonMonitor can be seen here:

https://github.com/scrapinghub/spidermon/blob/a7e195f951a4b9a4837943dff0b2ae727b76f237/spidermon/contrib/scrapy/monitors/monitors.py#L557-L579

where `number_of_jobs`'s value is defined by the `SPIDERMON_JOBS_COMPARISON` setting.

## Issue

Supposedly, we should get the `SPIDERMON_JOBS_COMPARISON` number of previous jobs to compute the average, as stated in the docs, but we might get more because of how those jobs are iterated.

## Debugging

* The first `_jobs = client.spider.jobs.list(..., count=number_of_jobs)` call will get `number_of_jobs` at most, as it is expected.
* The function then keeps iterating, skipping 1000 jobs ahead (start+= 1000) to fetch `#number_of_jobs` more.
    * That 1000 number might come from the fact that `client.spider.jobs.list()` doesn't allow the return of more than 1000 jobs, so the intention might have been to paginate the results to get everything. But because we only want `#number_of_jobs` (which should be less than 1000, otherwise the function breaks), then this causes us to fetch more jobs than requested instead.
* It does so until there are no more jobs available to list

For example, if some spider has 3500 jobs that meet the requested criteria and SPIDERMON_JOBS_COMPARISON is set to 10, then this will happen:
* `Round #0`: len(_jobs) == 10 (client.spider.jobs.list(start=0, count=10))
* `Round #1`: len(_jobs) == 20 (client.spider.jobs.list(start=1000, count=10))
* `Round #2`: len(_jobs) == 30 (client.spider.jobs.list(start=2000, count=10))
* `Round #3`: len(_jobs) == 40 (client.spider.jobs.list(start=3000, count=10))
* `Round #5`: stop (client.spider.jobs.list(start=4000, count=10))

Resulting in 40 previous jobs, when only ten were requested.

## Expected Behaviour

Only SPIDERMON_JOBS_COMPARISON number of jobs should be fetched as previous jobs, and those should be the latest ones that meet the requested criteria.

PR #444, opened 2024-05-03
Fix get_jobs method in Comparison Monitor

VMRuiz:
Fixes https://github.com/scrapinghub/spidermon/issues/442

### DIFF
--- a/spidermon/contrib/scrapy/monitors/monitors.py
+++ b/spidermon/contrib/scrapy/monitors/monitors.py
@@ -560,10 +560,11 @@ class ZyteJobsComparisonMonitor(BaseStatMonitor):
         total_jobs = []
         start = 0
         client = Client(self.crawler.settings)
+        MAX_API_COUNT = 1000
 
         while True:
             # API has a 1000 results limit
-            count = min(number_of_jobs - len(total_jobs), 1000)
+            count = min(number_of_jobs - len(total_jobs), MAX_API_COUNT)
             current_jobs = client.spider.jobs.list(
                 start=start,
                 state=states,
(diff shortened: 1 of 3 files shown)


VMRuiz:
Force pushed branch again to remove incorrectly staged files

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -557,26 +557,30 @@ def run(self, result):
...
-            _jobs = client.spider.jobs.list(
+        while True:
+            # API has a 1000 results limit
+            count = min(number_of_jobs - len(total_jobs), MAX_API_COUNT)
+            current_jobs = client.spider.jobs.list(
                 start=start,
                 state=states,
-                count=number_of_jobs,
+                count=count,
                 filters=dict(has_tag=tags) if tags else None,
```
curita, line 572: **issue (non-blocking):** This was already like this, so we might need to update it in another PR, but `filters=dict(has_tag=tags) if tags else None` doesn't do anything, at least in the version of `scrapinghub` I'm using (`2.4.0`). `has_tag=tags` does filter the jobs, though I'm not sure if that's backwards compatible.

curita (approved):
Changes look good to me!

Thanks for addressing these and the CI issue 🙇‍♀️

spidermon/contrib/scrapy/monitors/monitors.py
VMRuiz: Is there any other issues about this?

VMRuiz: Is there any other issues about this?

curita: No that I know of, no! And that issue is not related to these changes. We should be good to merge 👍

Issue #445, opened 2024-05-06
Scrapinghub jobs.list method called with incorrect filter parameter


VMRuiz:
`filters=dict(has_tag=tags) if tags else None` doesn't do anything, at least in the version of `scrapinghub` I'm using (`2.4.0`). `has_tag=tags` as a parameter by itself, without the `filters` wrap, does filter the jobs, though I'm not sure if that's backward compatible.

_Originally posted by @curita in https://github.com/scrapinghub/spidermon/pull/444#discussion_r1591016753_

PR #446, opened 2024-05-06
Fixes Scrapinghub jobs.list Method Called with Incorrect Filter Parameter

VMRuiz:
Fixes https://github.com/scrapinghub/spidermon/issues/445


Gallaecio (approved):

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -596,7 +596,7 @@ def _get_tags_to_filter(self):
             return {}
```
curita, line 596: **suggestion (non-blocking):** We might want to return empty lists for these two empty cases, though we don't really use them (as None is passed to the query instead).

curita (approved):
LGTM! Thanks for addressing it so quickly 🙏

PR #454, opened 2024-08-22
[WIP]Add New Monitors to track Requests Success Rate

shafiq-muhammad:
Added following monitors to test requests success rate
- ZyteAPIErrorRatioMonitor
- SPMErrorRatioMonitor

### DIFF
--- a/spidermon/contrib/scrapy/monitors/monitors.py
+++ b/spidermon/contrib/scrapy/monitors/monitors.py
@@ -332,9 +332,7 @@ class SPMErrorRatioMonitor(BaseScrapyMonitor):
 
     stat_name = "zyte_smartproxy/response/error"
 
-    @monitors.name(
-        "The error ratio should be less than threshold"
-    )
+    @monitors.name("The error ratio should be less than threshold")
     def test_maximum_error_ratio(self):
         if self.stat_name not in self.stats:
             message = f"Unable to find '{self.stat_name}' in job stats."
@@ -346,9 +344,11 @@ class SPMErrorRatioMonitor(BaseScrapyMonitor):
 
         errors = self.stats[self.stat_name]
         requests = self.stats["zyte_smartproxy/request"]
-        ratio = errors/requests
+        ratio = errors / requests
 
-        msg = "Too many failed SPM responses ({}) reached the maximum retry amount".format(errors)
+        msg = "Too many failed SPM responses ({}) reached the maximum retry amount".format(
+            errors
+        )
         self.assertLessEqual(ratio, threshold, msg=msg)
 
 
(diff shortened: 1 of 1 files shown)


spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -297,6 +299,59 @@ def test_minimum_successful_requests(self):
...
+
+    @monitors.name("The error ratio should be less than threshold")
+    def test_maximum_error_ratio(self):
+        if self.stat_name not in self.stats:
+            message = f"Unable to find '{self.stat_name}' in job stats."
+            self.skipTest(message)
+
+        threshold = self.crawler.settings.getint(SPIDERMON_MAX_SPM_ERROR_RATIO, -1)
+        if threshold < 0:
+            return
```
curita, line 343: **suggestion:** I think in this case we should notify that the monitor is not configured. Either via a regular log message, or by skipping the test with a message saying that `SPIDERMON_MAX_SPM_ERROR_RATIO` is not set.

Otherwise this monitor will be marked as "OK", when it didn't really run, which could make debugging harder.

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -297,6 +299,59 @@ def test_minimum_successful_requests(self):
...
+
+        errors = self.stats[self.stat_name]
+        requests = self.stats["zyte_smartproxy/request"]
+        ratio = errors / requests
```
curita, line 347: **todo:** We should handle the case where `requests` is zero, otherwise this will throw a ZeroDivisionError.

We could consider that case to have a `ratio` of zero, or we can skip it otherwise.

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -297,6 +299,59 @@ def test_minimum_successful_requests(self):
...
+
+        msg = "Too many failed SPM responses ({}) reached the maximum retry amount".format(
+            errors
+        )
+        self.assertLessEqual(ratio, threshold, msg=msg)
```
curita, line 352: **suggestion:** I think for both monitors we might want to have some condition for when to start checking, as spiders could have a high error rate at the start, but stabilize later.

For the alerts in the delivery-alerts channel, we only receive them after a certain amount of requests have been issued by the job (Reference: https://docs.google.com/document/d/1LBNgyXwx-w3eaHikCxYheKFcpMP0QfYMcnqGWT7BkPI/edit?tab=t.0#heading=h.rmcwqn39xzyn).

We could do something similar here and set a configurable minimum number of requests before these monitors start raising alerts.

spidermon/contrib/scrapy/monitors/monitors.py
```diff
@@ -297,6 +299,59 @@ def test_minimum_successful_requests(self):
         self.assertGreaterEqual(requests, threshold, msg=msg)
 
 
+@monitors.name("Zyte API Error Ratio monitor")
+class ZyteAPIErrorRatioMonitor(Monitor, BaseStatMonitor):
```
curita, line 303: **todo:** We'll have to add unitttests for the new monitor checks.

curita (changes requested):
Thanks Muhammad for these changes!

I pointed some few things to update, but these monitors are nearly done already :)

I'll be tagging @vineettomar06 for any further feedback about these monitors, and also so they can serve as a base for the new 429 error monitors.

curita:
Ah, didn't notice these were issued here instead of the common monitors repo.

We might want to submit them to the common monitors repo for the time being @shafiq-muhammad, while we finalize the repo with the Zyte product monitors.

PR #455, opened 2024-09-18
[actions][sentry] apply fingerprint rules to group issues better

gutsytechster:
Sentry groups issues by the provided message argument in the implemented scenario. However, when the message includes a pattern recognizable by sentry defined here - https://github.com/getsentry/sentry/blob/master/src/sentry/grouping/parameterization.py#L56, values against these patterns are replaced with placeholders and aren't used to group issues.

E.g., if the sentry title contains hostnames with the same exception, the two would be grouped as below

Prashant Test Local | Development | Spider bookspider.com notification
Prashant Test Local | Development | Spider quotespider.com notification

grouped into

Prashant Test Local | Development | Spider `<hostname>` notification

This would be unintended behaviour, and we'd like to have the two spiders raise two different sentry alerts.

Providing spider's title to the fingerprint would ensure that the exact message is used to create fingerprints, and hence the two issues will have separate fingerprint values and eventually two different sentry issues.

### DIFF
--- a/spidermon/contrib/actions/sentry/__init__.py
+++ b/spidermon/contrib/actions/sentry/__init__.py
@@ -136,12 +136,13 @@ class SendSentryMessage(Action):
             scope.set_extra("failed_monitors", message.get("failed_monitors", []))
 
             sentry_client.capture_event(
-                {
+                event={
                     "message": "{title} \n {description}".format(
                         title=message.get("title"),
                         description=message.get("failure_reasons", ""),
                     ),
                     "level": self.sentry_log_level,
+                    "fingerprint": [message.get("title")],
                 },
                 scope=scope,
             )


muzaffaryousaf:
@gutsytechster were you able to test the changes ? Thanks.

gutsytechster:
Hi @muzaffaryousaf 

Yes. Here's a [sentry alert](https://scrapinghub-maintenance.sentry.io/issues/5865168245/events/?project=5028633&referrer=issue-stream&statsPeriod=90d&stream_index=4) that grouped events of the different domain alike spiders together. This was previous to fixes.

Here's how it grouped these events
image

After fixes, the two separate events were triggered for these two spiders - [bookspider.com](https://scrapinghub-maintenance.sentry.io/issues/5869296401/?query=Prashant&referrer=issue-stream&statsPeriod=90d&stream_index=1) and [quotespider.com](https://scrapinghub-maintenance.sentry.io/issues/5869342495/?query=Prashant&referrer=issue-stream&statsPeriod=90d&stream_index=0)

Here's how it grouped these events.
image
image

Notice `Fingerprint values` in these images, where spider title is used to group on top of the `{{default}}` values.

gutsytechster:
After discussion with monitoring team, we'd like to categorize only based on the event title only, discarding the default grouping. It'll be unique for each spider and it will group all types of spidermon errors under the same issue.

VMRuiz (approved):

PR #458, opened 2024-11-11
Update `setup.py`

curita:
Made some changes in `setup.py` to add some extra details to [Spidermon's PyPI page](https://pypi.org/project/spidermon/).

1. Used the existing README as the `Project description` of the library
2. Added "Documentation", "Source", "Tracker" and "Release notes" as additional `Project links`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,7 @@ setup(
     author_email="opensource@zyte.com",
     description=("Spidermon is a framework to build monitors for Scrapy spiders."),
     long_description=open("README.rst", encoding="utf-8").read(),
+    long_description_content_type="text/x-rst",
     license="BSD",
     packages=find_packages(),
     package_data={"spidermon": ["VERSION"]},
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,12 @@ setup(
     name="spidermon",
     version="1.23.0",
     url="https://github.com/scrapinghub/spidermon",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -4,10 +4,16 @@
...
+        "Documentation": "https://spidermon.readthedocs.io/",
+        "Source": "https://github.com/scrapinghub/spidermon",
+        "Tracker": "https://github.com/scrapinghub/spidermon/issues",
+        "Release notes": "https://spidermon.readthedocs.io/en/latest/changelog.html",
+    },
     author="Zyte",
     author_email="opensource@zyte.com",
     description=("Spidermon is a framework to build monitors for Scrapy spiders."),
-    long_description=("Spidermon is a framework to build monitors for Scrapy spiders."),
+    long_description=open("README.rst", encoding="utf-8").read(),
```
Gallaecio, line 16: We are missing it in Scrapy, but you should also set `long_description_content_type="text/x-rst"`

Gallaecio (approved):

setup.py
wRAR: That's why I just filed #459

curita: I saw it in other projects but wasn't sure if it was necessary! I just added it to the PR.

curita:
Thank you!! ☺️

PR #462, opened 2025-04-04
Use SHUB_VIRTUAL_SPIDER as spider name when available

fcanobrash:


spidermon/contrib/scrapy/extensions.py
```diff
@@ -238,5 +240,6 @@ def _generate_data_for_spider(self, spider):
             else [],
             "crawler": self.crawler,
             "spider": spider,
+            "sc_spider_name": os.getenv("SHUB_VIRTUAL_SPIDER", spider.name),
```
VMRuiz, line 243:
```suggestion
            "sc_spider_name": os.getenv("SHUB_VIRTUAL_SPIDER") or spider.name,
``` 

Otherwise `SHUB_VIRTUAL_SPIDER=""` will cause `sc_spider_name` to be `""`

spidermon/contrib/stats/statscollectors/local_storage.py
```diff
@@ -9,7 +9,8 @@
 class LocalStorageStatsHistoryCollector(StatsCollector):
     def _stats_location(self, spider):
         statsdir = data_path("stats", createdir=True)
-        return os.path.join(statsdir, f"{spider.name}_stats_history")
+        spider_name = os.getenv("SHUB_VIRTUAL_SPIDER", spider.name)
```
VMRuiz, line 12: Refactor spider name extraction into a utility function that can be reused across all methods. 
That way we can introduce a setting to change it for a different one if wanted.

VMRuiz (changes requested):
This is a backward incompatible change. We should include an explanation somewhere on the docs.

VMRuiz (approved):

PR #465, opened 2025-12-11
[DC-1071] Field coverage where field is set to default value

NiltonGMJunior:
This PR aims to solve the issue set by [DC-1071](https://zyte.atlassian.net/jira/software/c/projects/DC/boards/543?selectedIssue=DC-1071), from the monitoring chapter.

The issue states that default/falsy values can distort the real field coverage. For example, a value of 0, "N/A", "TBD", "" (empty string), could be considered valid and count towards the field coverage, which is often undesirable. 

We already have a config that allows users to skip a field when it's None/null (as oposed to being not present in the item) - SPIDERMON_FIELD_COVERAGE_SKIP_NONE.

This PR proposes two new fields:
SPIDERMON_FIELD_COVERAGE_SKIP_FALSY - skips Python's falsy values, set by the next setting
SPIDERMON_FIELD_COVERAGE_SKIP_VALUES - allows the user to set a list of default values to be skipped when counting values for field coverage.

It's also worth noting that SPIDERMON_FIELD_COVERAGE_SKIP_FALSY defaults to True and the default SPIDERMON_FIELD_COVERAGE_SKIP_VALUES define the falsy value project-wide as: '' (empty string), [], {}, 'N/A' and '-'. This list does NOT include None, which is handled by the existing SPIDERMON_FIELD_COVERAGE_SKIP_NONE setting.

[DC-1071]: https://zyte.atlassian.net/browse/DC-1071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lucasfrancoqueiroz2 (changes requested):
@NiltonGMJunior the code looks good to me. I just have too points:

- I believe the default value of SPIDERMON_FIELD_COVERAGE_SKIP_FALSY should be True and we should set as default the most common empty placeholders for SPIDERMON_FIELD_COVERAGE_SKIP_VALUES. This is a conservative approach that will force the developer to consciously change the settings if needed and avoids the issue to be hidden until being discovered.
- It seems some checks were not successful. Can you check?

lucasfrancoqueiroz2 (approved):

NiltonGMJunior:
Fixed merge conflicts and blacken-docs issue (updating docs only)

docs/settings.rst
```diff
@@ -189,6 +189,132 @@ If this setting is not provided or set to ``False``, spider statistics will be:
         "spidermon_item_scraped_count/dict/field_2": 1,
     }
 
+SPIDERMON_FIELD_COVERAGE_SKIP_FALSY
+-----------------------------------
+Default: ``True``
```
lucasfrancoqueiroz2, line 194: @NiltonGMJunior maybe this comment is outside of the PR scope, but I think if the default value for SPIDERMON_FIELD_COVERAGE_SKIP_FALSY is True (which I think is correct) the default value for SPIDERMON_FIELD_COVERAGE_SKIP_NONE should also be True.

docs/settings.rst
```diff
@@ -189,6 +189,132 @@ If this setting is not provided or set to ``False``, spider statistics will be:
...
+If this setting is set to ``True``, spider statistics will be:
+
+.. code-block:: python
+
+    {
+        "spidermon_item_scraped_count/dict": 2,
+        "spidermon_item_scraped_count/dict/field_1": 1,  # Ignored empty string
+        "spidermon_item_scraped_count/dict/field_2": 2,
+        "spidermon_item_scraped_count/dict/field_3": 1,  # Ignored empty list
+        "spidermon_item_scraped_count/dict/field_4": 1,  # Ignored zero
```
lucasfrancoqueiroz2, line 226: @NiltonGMJunior I am wondering if 0 shouldn't have a special treatment although it is Falsy in the Python. In the scraping context, usually when a field has a 0 value it was correctly extracted otherwise it would have None or an empty list in general.

lucasfrancoqueiroz2 (approved):
Approved with comments.

AdrianAtZyte (approved):

PR #475, opened 2026-04-08
feat: allow SPIDERMON_DICT_FIELDS_COVERAGE_LEVELS to be configured per field

kishan3:
Closes #466


AdrianAtZyte (approved):
Looks good to me. I wonder if we should use some other key for the "default" that is less likely to match a real field name, e.g. `"*"`, but no strong opinion on that.

kishan3:
> Looks good to me. I wonder if we should use some other key for the "default" that is less likely to match a real field name, e.g. `"*"`, but no strong opinion on that.

@copilot let's do this and update the PR.

AdrianAtZyte:
@kishan3 Do you need to address the CI issues?

kishan3:
> @kishan3 Do you need to address the CI issues?

@copilot can you check CI issues?

kishan3:
> > @kishan3 Do you need to address the CI issues?
> 
> @copilot can you check CI issues?

@AdrianAtZyte just to note this is existing issue as per agent. Below is from copilot. 

> So the smtp.py and test files are the same on master. The CI failure is a pre-existing issue introduced by the merge with master (the merge commit 10e7d3b which merged with master). In master, Scrapy 2.15.0 deprecated scrapy.mail, which causes the test collection to fail.

AdrianAtZyte:
Can you undo the unrelated changes that Copilot introduced, to be handled as part of https://github.com/scrapinghub/spidermon/issues/477 (and probably not the way Copilot did here).

kishan3:
> Can you undo the unrelated changes that Copilot introduced, to be handled as part of #477 (and probably not the way Copilot did here).

Thanks @AdrianAtZyte reverted now.

PR #481, opened 2026-08-25
Replace boto (v2) with boto3 in S3 report action

### LINKED ISSUE #480: Proposal: drop the boto (v2) dependency
Spidermon still ships `boto>=2.32.1` in the `monitoring` extra, but it's used in exactly one place: `spidermon/contrib/actions/reports/s3.py` (`S3Connection` + `Key`). Everything else AWS-related — `SES` email and `SNS` actions — was already migrated to `boto3` back in #285.

Why remove it:

1. `boto` is broken on `Python 3.12`, `3.13` and `3.14` — three of the five versions we advertise as supported.
2. It's archived and deprecated. Last `boto` release is `2.49.0` from July 2018. No security fixes, and no compatibility statement for the Python versions we claim to support (`3.10`–`3.14`). Some of its modules still import stdlib modules that no longer exist (e.g. `cgi`, removed in `3.13`).
3. We already depend on `boto3`. Porting `S3Uploader` to `boto3` (`client.put_object` / `client.upload_fileobj`) removes a dependency instead of adding one, and gives us `SigV4` signing by default — `boto` v2 defaults to `SigV2`, which AWS no longer supports in newer S3 regions, so `CreateS3Report` is likely already broken for anyone on a recent bucket/region.

Suggested plan: rewrite `S3Uploader` on top of `boto3`, keep the public `CreateS3Report` API unchanged, drop `boto` from `pyproject.toml`, and add the tests the module never had.

AdrianAtZyte:
Resolves https://github.com/scrapinghub/spidermon/issues/480

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ Tracker = "https://github.com/scrapinghub/spidermon/issues"
 
 [project.optional-dependencies]
 monitoring = [
-    "boto>=2.32.1",
     "boto3>=1.10.0",
     "itemadapter>=0.1.0",
     "jinja2>=3.0.0",
--- a/spidermon/contrib/actions/reports/s3.py
+++ b/spidermon/contrib/actions/reports/s3.py
@@ -1,8 +1,7 @@
 import secrets
 from pathlib import Path
 
-from boto.s3.connection import S3Connection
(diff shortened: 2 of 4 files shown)


VMRuiz (approved):
LGTM

PR #485, opened 2026-08-26
Degrade gracefully when scrapy.cfg can't be found

### LINKED ISSUE #200: Unable to find scrapy.cfg file to infer project data dir
image
i got error message like this when deployed scrapy project to scrapyd, even when scrapy.cfg is included in the egg file

I have deployed a scrapy project to scrapyd, but I think there is a problem with the spidermon, because without scrapyd it's fine

AdrianAtZyte:
Resolves https://github.com/scrapinghub/spidermon/issues/200


VMRuiz:
Why would you change the default `STATS_CLASS` from Scrapy to Spidermon's `LocalStorageStatsHistoryCollector` to persists stats and then have stats persistence disabled due to incorrectly configured project? 

Seems like a invalid configuration issue that the user should solve by either enabling `scrapy.cfg` in their project or using a different `STATS_CLASS` like the default one provided by Scrapy.

AdrianAtZyte:
You have a point. In which case, I wonder if we should just close this PR and close the issue as well as won’t fix. Or create a PR that improves the error message. Or move the issue to scrapyd, in case the root issue (not copying scrapy.cfg) should be fixed there.

VMRuiz:
Ideally: 
1. Log the error of why `LocalStorageStatsHistoryCollector` is not working as expected.
2. Move the missing scrapy.cfg file issue upstream to scrapyd.

VMRuiz (approved):

PR #487, opened 2026-08-26
Log monitor failures at their configured level instead of always ERROR

### LINKED ISSUE #396: Support different severity levels for Spidermon alerts
## Background

All Spidermon failed checks are reported as ERRORs at the moment. Everything belongs to the same category and can't be addressed differently. It could be nice to have different severity levels to signal how critical those failed checks are and allow us to take distinct actions based on them.

This idea was initially introduced in the discussions for https://github.com/scrapinghub/spidermon/issues/375.

## Proposal

We could use existing log levels as severity levels and log those messages with that severity. ERROR could remain the default severity and introduce a lower priority (WARNING) and a higher priority (CRITICAL) severity level.

Initially, this could be supported only in custom monitors via some new parameter in the assertions of the test methods. And we could support assigning different Spidermon Actions based on these severity levels.

This feature should be backward compatible ideally, considering ERRORs the default severity.

AdrianAtZyte:
Resolves https://github.com/scrapinghub/spidermon/issues/396


VMRuiz (commented):
Is there any docs about configuring level for monitors?

VMRuiz (approved):

PR #496, opened 2026-08-26
Report each unexpected JSON schema field as its own stat entry

### LINKED ISSUE #227: no proper message or monitors for extra fields when using jsonschema
The stats for unexpected fields is almost unreadable and can't be used for writing a monitor to raise error if a crawler is crawling any unexpected field.

`'spidermon/validation/fields/errors/unexpected_fields_upopularity_ranking_uratings_utotal_competition_ucertificate_of_excellence_unum_of_photos_uclaimed_uprice_range_were_unexpected': 1,
 'spidermon/validation/fields/errors/unexpected_fields_upopularity_ranking_uratings_utotal_competition_ucertificate_of_excellence_unum_of_photos_uclaimed_uprice_range_were_unexpected/': 1,`

AdrianAtZyte:
Resolves https://github.com/scrapinghub/spidermon/issues/227


VMRuiz:
Not sure whether the right approach is:

1. One generic stat entry counting all unexpected fields, or
2. One custom stat entry per unexpected field

For monitoring purposes, option 1 is sufficient. Option 2 risks bloating the stats payload if too many distinct entries get added, which can cause issues with the Scrapy Cloud stats monitor.

Is there a real use case for having option 2 instead of option 1?

AdrianAtZyte:
> Is there a real use case for having option 2 instead of option 1?
>
> We get both either way: `add_field_error` always increments `fields/errors/unexpected_field` as well as `fields/errors/unexpected_field/<field>`, so option 1 is the aggregate key and option 2 is the same per-field key every other error type already produces, e.g. `missing_required_field/<field>`.
>
> The current behaviour is actually the bloated one: the whole `('a', 'b' were unexpected)` string becomes the stat key, so every distinct combination of unexpected fields creates a new key (plus a duplicate ending in `/` because the field name is empty). This bounds it to one key per unexpected field, and lets a monitor alert on a specific field name, which is what #227 asks for.

I do wonder if we should consider making some of these stats optional, opt-in even.

PR #500, opened 2026-08-26
Add FeedExportMonitor for feed export failures

### LINKED ISSUE #196: Add monitors for export validations
Currently, there are monitors for item count validation, checking errors, etc. There is one important monitor that is missing, to monitor whether the data has been exported to FEED_STORAGE or not. 

This feature is depended and can be added after https://github.com/scrapy/scrapy/issues/3947

AdrianAtZyte:
Resolves https://github.com/scrapinghub/spidermon/issues/196


VMRuiz (changes requested):
Can you add a test for the case where the expected stats are missing? Some monitors fail when they can’t find the expected stat, even though they don’t necessarily need to fail in that scenario.

VMRuiz (approved):

PR #503, opened 2026-08-27
Make SPIDERMON_MONITOR_SKIPPING_RULES support any suite and per-method skipping

### LINKED ISSUE #339: skip_if decorator for monitors
This came from a real use case:

`FieldCoverageMonitor` requires that we have at least one item returned, or else it will fail. However there are situation when we allow a spider to return no items. Think about a real estate website searching for locations in a certain neighborhood, sometimes even with the right code, we don't have any result, so we should not fail our monitors.

This is not possible with our built-in monitors, and the solution is create a new custom monitor, base in a built-in monitor that implements this check before doing the test.

`pytest` library has a `skipif` decorator (https://docs.pytest.org/en/6.2.x/skipping.html#id1) where you can set some rules and if these rules are not met, the test are not executed. We could consider to add something like that in Spidermon so we can decide if we want to execute or skip a test considering a condition (check a stat value could be enough).

Some options to accomplish that:
- Create a `skipif` decorator that can be applied to a `MonitorSuite` class, so the entire suite will not be executed whether the condition is not met. This looks simpler, as we can use our existing built-in monitors as-it-is now and the control is 
- Create a `skipif` decorator for the `test_` methods, so that method is not executed whether the condition is met. This has the drawback that when we want to use a built-in monitor, we will need to create a custom monitor just to add the decorator

Opening this issue to get more ideas and suggestions if this is a good idea and if so, what is the best solution for that.

AdrianAtZyte:
https://github.com/scrapinghub/spidermon/issues/339#issuecomment-5431839315

Resolves https://github.com/scrapinghub/spidermon/issues/339.


docs/settings.rst
```diff
@@ -535,6 +535,9 @@ Default: ``None``
 
 A dictionary where keys represent the names of the monitors to be skipped, and the corresponding values are lists containing either method names or lists defining skip conditions.
 
+.. versionchanged:: VERSION
```
curita, line 538: Non-blocking: this note covers the per-method key format, but the PR also changes *where* the rules apply. Previously they only took effect when the suite was `SpiderCloseMonitorSuite` or a subclass of it; now they apply to any `BaseScrapyMonitor` in any suite (including periodic and spider-open suites). A user who already has `SPIDERMON_MONITOR_SKIPPING_RULES` configured but runs a custom suite will see monitors start being skipped after upgrading. It would be good to mention that here as well, and to add a release notes entry to `CHANGES.rst` covering both changes.

docs/settings.rst
```diff
@@ -535,6 +535,9 @@ Default: ``None``
...
+    A key can also target a single test method of a monitor, instead of every method in it, using the ``"<monitor name>/<method name>"`` format.
```
curita: Non-blocking: it may be worth clarifying which names are matched. Both parts of the key must use the display names, i.e. the names set with `@monitors.name(...)` when present (this is what `Monitor.name` returns), falling back to the class/method identifiers otherwise. For example, for `FinishReasonMonitor` the key is `"Finish Reason Monitor/Should have the expected finished reason(s)"`, and the Python identifier `"Finish Reason Monitor/test_should_finish_with_expected_reason"` does not match. Since several built-in monitors decorate their test methods, this is easy to get wrong.

spidermon/contrib/scrapy/monitors/base.py
```diff
@@ -34,13 +34,12 @@ def run(self, result):
...
-            and self.monitor_name
-            and self.skip_rules.get(self.monitor_name)
-        ):
-            skip_rules = self.skip_rules[self.monitor_name]
-            for rule in skip_rules:
+        crawler = self.data.get("crawler")
+        if not crawler:
+            return False
+        skip_rules = crawler.settings.get("SPIDERMON_MONITOR_SKIPPING_RULES") or {}
+        for key in (self.name, self.monitor_name):
```
curita, line 41: Non-blocking: `run()` above still logs `Skipping {self.monitor_name} monitor`. With a per-method rule, skipping only e.g. `"MyMonitor/test_two"` logs a message that implies the whole monitor was skipped. Logging `self.name` instead would be accurate for both monitor-level and method-level skips.

curita (approved):
LGTM. The stateless run-time lookup is a clear improvement over injecting `skip_rules` attributes at suite init time, and it fixes the undocumented limitation that the rules only worked with `SpiderCloseMonitorSuite`. I verified the new tests pass, the full test suite shows no regressions compared to master, and per-method/monitor-level/no-crawler cases all behave as documented.

A few non-blocking suggestions inline.

curita:
Looks ready to merge to me, thanks Adrian 🙇‍♀️

